PR #13, opened 2022-05-16
MCS Testing 

india-kerle:
Hey Chris - thanks for taking a look at this! This pretty much covers most scripts with the exception of the process_mcs_installers.py and process_mcs_utils.py. Hopefully this is okay! As I had mentioned, I think it would be easiest if Alex sends our installers vs. installations data in separate files so the testing script can work as it should.


asf_core_data/config/base.yaml
```diff
@@ -1,3 +1,9 @@
+MCS_RAW_S3_PATH: "inputs/MCS/mcs_heat_pumps.xlsx"
+MCS_RAW_LOCAL_PATH: "/inputs/data/MCS/mcs_heat_pumps.xlsx"
+MCS_RAW_INSTALLER_S3_PATH: "inputs/MCS/mcs_installer_information.xlsx"
+MCS_RAW_INSTALLER_CONCAT_S3_PATH: "inputs/MCS/mcs_installers.csv"
+lOCAL_NEW_MCS_DATA_DUMP_DIR: "/inputs/data/mcs/"
```
ch-williamson: Lowercase L at the start of this one

asf_core_data/getters/data_getters.py
```diff
@@ -27,11 +27,17 @@ def load_s3_data(s3, bucket_name, file_name):
     bucket_name: The S3 bucket name
     file_name: S3 key to load
     """
+    obj = s3.Object(bucket_name, file_name)
```
ch-williamson: I don't think this is needed (it's used in save_to_s3 because that deals with pickles and jsons)

ch-williamson: This also means that the s3 parameter isn't needed

asf_core_data/pipeline/mcs/generate_mcs_data.py
```diff
@@ -27,6 +27,7 @@
     "MCS_INSTALLATIONS_EPC_MOST_RELEVANT_PATH"
 ]
 installations_raw_s3_path = config["INSTALLATIONS_RAW_S3_PATH"]
+installers_raw_s3_path = installers_raw_s3_path["MCS_RAW_INSTALLER_CONCAT_S3_PATH"]
```
ch-williamson: Should be `config[...`

asf_core_data/pipeline/mcs/generate_mcs_data.py
```diff
@@ -37,24 +38,30 @@
...
     """
 
     bucket = s3.Bucket(bucket_name)
 
-    installations_keys_and_dfs = [
+    keys_and_dfs = [
         (object.key, load_s3_data(s3, bucket_name, object.key))
         for object in bucket.objects.filter(Prefix=raw_data_s3_folder)
-        if "installations" in object.key
+        if "installations" or "installers" in object.key
```
ch-williamson: I think this might need to be `if ("installations" or "installers") in object.key`, as currently this parses as `"installations" or ("installers" in object.key)` and is treated as True

asf_core_data/pipeline/mcs/generate_mcs_data.py
```diff
@@ -80,19 +106,33 @@ def concatenate_save_raw_installations():
                     )
                 )
 
-    installations_dfs = [key_and_df[1] for key_and_df in installations_keys_and_dfs]
+    installations_dfs = []
+    for key, dfs in keys_and_dfs:
+        if key:
+            if type(dfs) == dict:
+                installations_dfs.append(dfs[installations_key])
+            elif type(dfs) == DataFrame:
```
ch-williamson: I think this needs to be `pd.DataFrame`

asf_core_data/pipeline/mcs/generate_mcs_data.py
```diff
@@ -64,13 +71,32 @@ def concatenate_save_raw_installations():
...
+            installations_key = " ".join(
+                [
+                    key
+                    for key in (list(key_and_df[1].keys()))
+                    if "installs" in key.lower()
+                ]
+            )
+            installer_key = " ".join(
+                [key for key in (list(df.keys())) if "installer" in key.lower()]
+            )
```
ch-williamson: I'm not sure what this is doing - currently both strings are empty. Should this be outside the `for key_and_df in keys_and_dfs` loop? I think `installs` should be `installations` and `df` should be `key_and_df[1]` as well

asf_core_data/pipeline/mcs/generate_mcs_data.py
```diff
@@ -64,13 +71,32 @@ def concatenate_save_raw_installations():
...
+
+            # append quarterly installer data
+            installer_dfs.append(key_and_df[1][installer_key])
+
             if (
                 (
-                    pd.to_datetime(key_and_df[1]["Commissioning Date"]).dt.date
+                    pd.to_datetime(
+                        key_and_df[1][installations_key]["Commissioning Date"]
```
ch-williamson: `key_and_df[1]` is a DataFrame so I don't think this will work

asf_core_data/pipeline/mcs/process/process_mcs_utils.py
```diff
@@ -105,6 +105,49 @@ def clean_company_name(company_name):
...
+        data[hp_types.columns[0]] = (
+            hp_types.where(hp_types.ne(0), np.nan)
+            .bfill(axis=1)[hp_types.columns[0]]
+            .fillna(0)
+        )
+        data.drop(columns=hp_types.columns[1], axis=1, inplace=True)
+
+    data.columns = [
+        " ".join(c.split(" ")[:-1]) if "Installation" in c else c
+        for c in raw_installers_clean.columns
```
ch-williamson: `raw_installers_clean` isn't defined

asf_core_data/pipeline/mcs/process/process_mcs_utils.py
```diff
@@ -105,6 +105,49 @@ def clean_company_name(company_name):
...
+    data = data[[col for col in data.columns if "Design" not in col]]
+    # merge columns iteratively
+    for keyword in keywords_to_merge:
+        hp_types = data[[col for col in data.columns if keyword in col]]
+        data[hp_types.columns[0]] = (
+            hp_types.where(hp_types.ne(0), np.nan)
+            .bfill(axis=1)[hp_types.columns[0]]
+            .fillna(0)
+        )
+        data.drop(columns=hp_types.columns[1], axis=1, inplace=True)
```
ch-williamson, line 146: Here I get that 1 is out of bounds - hp_types is a single-column dataframe

asf_core_data/pipeline/mcs/process/process_mcs_utils.py
```diff
@@ -105,6 +105,49 @@ def clean_company_name(company_name):
...
+    # installation and design is the same; drop columns
+    data = data[[col for col in data.columns if "Design" not in col]]
+    # merge columns iteratively
+    for keyword in keywords_to_merge:
+        hp_types = data[[col for col in data.columns if keyword in col]]
+        data[hp_types.columns[0]] = (
+            hp_types.where(hp_types.ne(0), np.nan)
+            .bfill(axis=1)[hp_types.columns[0]]
+            .fillna(0)
+        )
```
ch-williamson, line 145: Is all of this needed? Not completely sure about the bfill either - is this imputing missing values with whatever's in the next record?

asf_core_data/pipeline/mcs/test/compare_mcs_installations.py
```diff
@@ -107,9 +103,13 @@ def within_mcs_installers_check(mcs_installers):
         }
     )
 
-    mcs_installer_comp_data = schema_withchecks.validate(
-        mcs_installers_no_unspecified, lazy=True
-    )
+    try:
+        mcs_installer_comp_data = schema_withchecks.validate(
```
ch-williamson: `mcs_installer_comp_data` doesn't seem to be used later

asf_core_data/pipeline/mcs/test/gdrive_to_s3.py
```diff
@@ -0,0 +1,76 @@
...
+    """Pulls MCS data dumps from Google Drive to a local dir and s3.
+    
+    Inputs:
+        local_data_dump_dir (str): local path to store MCS data dumps.
+        s3_data_dump_dir (str): path in s3 to store MCS data dumps. 
+
+    """
+    # get all folder names
+    all_folders_list = drive.ListFile(
+        {"q": "'root' in parents and trashed=false"}
```
ch-williamson: For me this just returns everything in my personal drive, so the data dumps folder isn't there... any way of providing this with a permanent link to the folder? Or moving the folder into the ASF team drive and using an approach like [this](https://stackoverflow.com/questions/55729923/how-can-i-access-a-team-drive-instead-of-personal-google-drive-using-pydrive)?

asf_core_data/pipeline/mcs/test/gdrive_to_s3.py
```diff
@@ -0,0 +1,76 @@
...
+    ).GetList()
+    # get id of MCS Data Dumps folder name
+    mcs_folder_id = [
+        folder["id"]
+        for folder in all_folders_list
+        if folder["title"] == "MCS Data Dumps"
+    ]
+    # get mcs files and parse created date
+    mcs_files = drive.ListFile(
+        {"q": "'{}' in parents and trashed=false".format("".join(mcs_folder_id))}
```
ch-williamson: Will the `.join` actually work if there's more than one folder with this name? I'd assume it would just make an invalid ID, so should this just be `format(mcs_folder_id[0])` (maybe with a warning that multiple folders were found)?

asf_core_data/pipeline/mcs/test/gdrive_to_s3.py
```diff
@@ -0,0 +1,76 @@
...
+    ).GetList()
+    installers = []
+    installations = []
+    for mcs_file in mcs_files:
+        time = mcs_file["createdDate"].replace("T", " ").replace("Z", "")
```
ch-williamson, line 36: It looks like the createdDate is a datetime (without T/Z) so I don't think the replacement is needed

asf_core_data/pipeline/mcs/test/gdrive_to_s3.py
```diff
@@ -0,0 +1,76 @@
...
+    for _ in (latest_installations, latest_installers):
+        name, ext = _["title"].split(".")[0], _["title"].split(".")[1]
+        timestamp = str(_["createdDate"]).split(" ")[0].replace("-", "_")
+        latest_dump_name = name + "_" + timestamp + "." + ext
+        output_path = str(PROJECT_DIR) + local_data_dump_dir + latest_dump_name
+        s3_path = s3_data_dump_dir + latest_dump_name
+        file_id = _["id"]
+        file = drive.CreateFile({"id": file_id})
+        print(f"downloaded file {file_id} locally...")
+        file.GetContentFile(output_path)
```
ch-williamson: Maybe switch these two so the confirmation happens after the download? Just in case the download fails but the confirmation still prints which could be confusing

asf_core_data/pipeline/mcs/test/mcs_test.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+python gdrive_to_s3.py
+latest_outputs=$(aws s3api list-objects --bucket asf-core-data --prefix inputs/MCS/latest_raw_data/ --output text --query "Contents[].{Key: Key}")
+latest_outputs_array=($latest_outputs)
+new_installations="${latest_outputs_array[1]}"
+new_installers="${latest_outputs_array[2]}"
```
ch-williamson, line 7: I don't have much bash experience, but just checking that this still works now that this folder will contain multiple installation/installer files rather than just the latest ones?

asf_core_data/pipeline/mcs/test/gdrive_to_s3.py
```diff
@@ -0,0 +1,76 @@
+########################################
```
ch-williamson, line 1: Great script, pydrive2 is a handy tool to know about :)

ch-williamson (commented):
Hello! I've reviewed everything that differs from 1_mcs_processing. Once this is merged in I'll review the rest on the 1_mcs_processing -> dev PR :)

asf_core_data/pipeline/mcs/generate_mcs_data.py
india-kerle: hmmm odd that they're empty - because the new dataset returned is a dictionary where the key is the tab name and the value is the pd.DataFrame, this is supposed to get the installations key and the installer key to be able to call the dataframe. I'll take a second look!

india-kerle: okay i see the problem in the end - it should be key_and_df[1].keys() instead of df - this then returns the keys for the tab name and the pd.DataFrame.

asf_core_data/pipeline/mcs/generate_mcs_data.py
india-kerle: hmmm this shouldn't be a dataframe, this should be a dict now - I wonder if we were using different datasets?

india-kerle: although extremely confusing name now that i've modified things - I'll make that change!

asf_core_data/pipeline/mcs/process/process_mcs_utils.py
india-kerle: hmmm it should be two columns that have identical information i.e. Ground/Water Source Heat Pump Installation and Ground/Water Source Heat Pump. I'm trying to merge the two columns with bfill then just drop one of the columns i.e. hp_types.columns[1] - i'm wondering again if that has to do with the dataset being used based on me vs. you!

PR #14, opened 2022-05-18
MCS processing

ch-williamson:
---

Checklist:

- [ ] I have tested the code


asf_core_data/getters/mcs/get_mcs_installations.py
```diff
@@ -0,0 +1,74 @@
...
+from asf_core_data import PROJECT_DIR, get_yaml_config
+
+from asf_core_data.getters.data_getters import load_s3_data
+
+config = get_yaml_config(PROJECT_DIR / "asf_core_data/config/base.yaml")
+
+BUCKET_NAME = config["BUCKET_NAME"]
+INSTALLATIONS_RAW_S3_PATH = config["INSTALLATIONS_RAW_S3_PATH"]
+INSTALLATIONS_RAW_LOCAL_PATH = config["INSTALLATIONS_RAW_LOCAL_PATH"]
+RAW_DATA_S3_FOLDER = config["RAW_DATA_S3_FOLDER"]
```
athrado, line 17: For use as a package, we'll have to transfer all the config information into base_config.py since packages don't include any data files - and that's the simplest way to adapt to that. After we're done reviewing, we can go through and change all config references at once. Doesn't have to happen right away, just mentioning it here already. 

It will probably look something like this, only minor adjustments needed.

```suggestion
from config import base_config

BUCKET_NAME = base_config.BUCKET_NAME
INSTALLATIONS_RAW_S3_PATH = base_config.INSTALLATIONS_RAW_S3_PATH
```

asf_core_data/getters/mcs/get_mcs_installations.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def get_raw_installations_data(
+    local_path=str(PROJECT_DIR) + INSTALLATIONS_RAW_LOCAL_PATH,
```
athrado, line 21: Similarly, we'll have to go through and adjust all data paths so they have the same format throughout. I recommend having a data path that would lead to the ASF core data in a local dir or directly to the S3 bucket, and then a relative path that leads to the specific file. That way we can easily pick whether we want to load it from S3 or local dir, and give the relative path as default values using the config file.

Happy to discuss and later go through an adjust once we're done reviewing.

asf_core_data/getters/mcs/get_mcs_installations.py
```diff
@@ -0,0 +1,74 @@
...
+):
+    """Get raw MCS HP installation data (both domestic and non-domestic)
+    with no processing other than renaming columns and setting correct dtypes.
+
+    Args:
+        local_path (str, optional): Local path to raw data.
+            Defaults to str(PROJECT_DIR) + INSTALLATIONS_RAW_LOCAL_PATH.
+        refresh (bool, optional): Whether or not to update the local copy
+            of the data from S3. If local_path does not exist, data will be
+            pulled from S3 even if refresh is False. Defaults to True.
```
athrado, line 33: Refresh is an Interesting option! Maybe also worth discussing how we'd like this to behave across the asf-core-data package.

asf_core_data/getters/mcs/get_mcs_installations.py
```diff
@@ -0,0 +1,74 @@
...
+    Returns:
+        DataFrame: Raw HP installation data.
+    """
+
+    if refresh or not os.path.exists(local_path):
+        hps = load_s3_data(BUCKET_NAME, INSTALLATIONS_RAW_S3_PATH, usecols=None)
+    else:
+        hps = pd.read_excel(local_path)
+
+    hps = hps.astype(
```
athrado, line 46: pd.read_excel() has a dtype keyword argument (you can pass it a dict). Maybe that's a bit more efficient/clean than changing the types after loading. Can make a difference with huge datasets as EPC but maybe not that relevant for MCS.

asf_core_data/getters/mcs/get_mcs_installations.py
```diff
@@ -0,0 +1,74 @@
...
+            "address_1": str,
+            "address_2": str,
+            "address_3": str,
+            "postcode": str,
+        }
+    )
+    hps["commission_date"] = pd.to_datetime(hps["commission_date"])
+    hps = (
+        hps.fillna({"address_1": "", "address_2": "", "address_3": ""})
+        .convert_dtypes()
```
athrado: Not quite sure why we need that, but maybe I misunderstand it's use.
Edit: ah wait, does it change the NaN type for categorical features?

asf_core_data/pipeline/mcs/process/process_mcs_installations.py
```diff
@@ -0,0 +1,136 @@
...
+from asf_core_data.getters.mcs.get_mcs_installations import get_raw_installations_data
+from asf_core_data.pipeline.mcs.process.process_mcs_utils import clean_company_name
+
+config = get_yaml_config(PROJECT_DIR / "asf_core_data/config/base.yaml")
+
+MAX_COST = config["MCS_MAX_COST"]
+CLUSTER_TIME_INTERVAL = config["MCS_CLUSTER_TIME_INTERVAL"]
+
+
+def add_columns(hps):
```
athrado: Maybe _add_hp_features_ to be a bit more specific?

asf_core_data/pipeline/mcs/process/process_mcs_installations.py
```diff
@@ -0,0 +1,136 @@
...
+        "flow_temp": "Flow Temp: ([^\|]+)",
+        "scop": "SCOP: ([^\)]+)",
+    }
+    for product_feat, regex in product_regex_dict.items():
+        hps[product_feat] = [
+            re.search(regex, product).group(1).strip() for product in hps["products"]
+        ]
+
+    # Add RHI field - any "Unspecified" values in rhi_status field signify
+    # that the installation is not for DRHI, missing values are unknown
```
athrado, line 43: I'm sure there's a reason behind that, but maybe explain why "unspecified" == False and missing == "unknown". Also why call it DRHI in the second line?

asf_core_data/pipeline/mcs/process/process_mcs_installations.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    hps["cost"] = hps["cost"].mask((hps["cost"] <= 0) | (hps["cost"] > max_cost))
+
+    hps["flow_temp"] = pd.to_numeric(hps["flow_temp"])
+    hps["flow_temp"] = hps["flow_temp"].mask(hps["flow_temp"] <= 0)
+
+    hps["scop"] = pd.to_numeric(hps["scop"].mask(hps["scop"] == "Unspecified"))
+    hps["scop"] = hps["scop"].mask(hps["scop"] == 0)
+
+    hps["capacity"] = hps["capacity"].mask(hps["capacity"] < 0)
```
athrado, line 74: Very clean and very clear, love it!

asf_core_data/getters/mcs/get_mcs_installations.py
```diff
@@ -0,0 +1,74 @@
...
+                "postcode",
+            ]
+        )
+        .reset_index(drop=True)
+    )
+
+    if verbose:
+        print("Shape of loaded data:", hps.shape)
+
+    return hps
```
athrado, line 73: Code runs without issues and very quick!
I would probably define a more precise path for where to save the output. If we drop everything directly in /outputs, it's going to be a bit messy. Ideally we have an MCS dir as we have an EPC dir in outputs.

asf_core_data/pipeline/mcs/process/process_mcs_installations.py
```diff
@@ -0,0 +1,136 @@
...
+    hps["diff_bwd"] = (
+        hps.sort_values(["postcode", "commission_date"])
+        .groupby("postcode")["commission_date"]
+        .diff()
+    )
+    hps["diff_fwd"] = (
+        hps.sort_values(["postcode", "commission_date"])
+        .groupby("postcode")["commission_date"]
+        .diff(periods=-1)
+    )
```
athrado, line 104: This is a really cool way of detecting mass retrofits. Well done, very efficient! Here a few comments:

a) Maybe add a few lines of comments explaining what exactly happens here. I had to try the code out myself to really undersand what .diff() does and why we need forward and backwards diffs. Maybe describe that you compute the distance between the commisioning date for consecutive installs at the same postcode. Now that I know how it works it seems obvious from the code, but it's really not easy to figure out. 

b) In practice, it's probably not a problem but I'm a bit worried that 31 days is a long time interval, especially since your method allows chaining up several intervals.

Is 30% mass retrofits really what we'd expect?

Have you sense checked that identified clusters share other characteristics such as same HP type or same installer company or similar costs or similar house types or construction age? Might be worth looking into at some point, also to check whether mass retrofits are mainly done on flats and terraced houses.

asf_core_data/pipeline/mcs/process/process_mcs_installations.py
```diff
@@ -0,0 +1,136 @@
...
+    installations_data = get_raw_installations_data(refresh=refresh)
+
+    installations_data = add_columns(installations_data)
+    installations_data = mask_outliers(installations_data)
+    installations_data = identify_clusters(installations_data)
+    # installations_data["installer_name"] = installations_data["installer_name"].apply(
+    #     clean_company_name
+    # )
+
+    return installations_data
```
athrado, line 139: Runs smoothly!

asf_core_data/pipeline/mcs/process/mcs_epc_joining.py
```diff
@@ -0,0 +1,392 @@
...
+        hps["postcode"].fillna("unknown").str.upper().str.replace(" ", "")
+    )
+
+    hps["standardised_address"] = [
+        # Make address 1 and 2 lowercase, strip whitespace,
+        # and combine into a single string separated by a space
+        remove_punctuation(add1).lower().strip()
+        + " "
+        + remove_punctuation(add2).lower().strip()
+        for add1, add2 in zip(hps["address_1"].fillna(""), hps["address_2"].fillna(""))
```
athrado: I think we did `fillna("")` for the address columns when loading, maybe not necessary anymore?

asf_core_data/pipeline/mcs/process/mcs_epc_joining.py
```diff
@@ -0,0 +1,392 @@
...
+                "ADDRESS1",
+                "ADDRESS2",
+                "POSTTOWN",
+                "POSTCODE",
+                "index_y",
+                "postcode_y",
+                "standardised_postcode_y",
+                "numeric_tokens_y",
+                "level_0",
+            ],
```
athrado: To avoid/reduce the x_y  columns, you could remove some of them before joining, e.g.

`.merge(epcs.reset_index()[`_only columns you really need_`]`

or `hps.drop(columns=[`_columns you like numeric tokens and others you drop later_`]`

Personally, I try to avoid creating columns with x and y tags if at all possible because I'm always worried I will confuse them one day.

asf_core_data/pipeline/mcs/process/mcs_epc_joining.py
```diff
@@ -0,0 +1,392 @@
...
+            "level_1": "original_epc_index",
+            "postcode_x": "postcode",
+        }
+    )
+
+    if drop_epc_address:
+        merged = merged.drop(columns="standardised_address_y")
+    else:
+        merged = merged.rename(columns={"standardised_address_y": "epc_address"})
+
```
athrado, line 264: Note sure where that would happen, but I'd love to have the UPRN from EPC in the joined mcs_epc file so I can easily match up the MCS installations with the EPC properties. Currently going through the epc_address, which is not so great.

asf_core_data/pipeline/mcs/process/mcs_epc_joining.py
```diff
@@ -0,0 +1,392 @@
...
+        hps = get_processed_installations_data()
+
+    if epcs is None:
+        epc_version = "preprocessed" if all_records else "preprocessed_dedupl"
+        print("Getting EPC data...")
+        # TODO: replace with line that loads most recent full EPC data
+        epcs = load_s3_data(
+            bucket_name,
+            "outputs/EPC/EPC_GB_preprocessed_and_deduplicated_sample_prox.csv",
+            usecols=None,
```
athrado: Let's resolve with using the asf-core-package.
I haven't implemented direct load from S3 but I will add that soon.

I assume it would look something like this:

```suggestion
        epcs = asf_core_data.getters.epc.epc_data.load_preprocessed_epc_data(data_path='S3', 
            version=epc_version,
            usecols=[relevant features you need for this, defined in base_config.py],
```

asf_core_data/pipeline/mcs/process/mcs_epc_joining.py
```diff
@@ -0,0 +1,392 @@
...
+        joined_df.loc[joined_df["last_epc_before_mcs"] | ~joined_df["epc_before_mcs"]]
+        .groupby("original_mcs_index")
+        .head(1)
+        .reset_index(drop=True)
+        .drop(columns=["epc_before_mcs", "last_epc_before_mcs"])
+    )
+
+    return filtered_data
+
+
```
athrado, line 378: Ended up not fully testing this script because I couldn't install recordlinkage. That's on my side, though. And I've tested this code in the past so I'm sure it runs fine. Will give it a full run again once we've integrated the final EPC data load.

asf_core_data/pipeline/mcs/process/mcs_epc_joining.py
```diff
@@ -0,0 +1,392 @@
...
+    filtered_data = (
+        joined_df.loc[joined_df["last_epc_before_mcs"] | ~joined_df["epc_before_mcs"]]
+        .groupby("original_mcs_index")
+        .head(1)
+        .reset_index(drop=True)
+        .drop(columns=["epc_before_mcs", "last_epc_before_mcs"])
+    )
+
+    return filtered_data
+
```
athrado, line 377: I think a main() would be good here that calls the functions join_mcs_epc_data() and  select_most_relevant_epc() after each other. This would both demonstrate how to use the functions and allow direct execution by just running this script.

athrado (commented):
Hi Chris, 
This looks really good. I especially loved the cluster detection, hadn't seen that before!
I think it only takes minor tweaks and then some adjustments to make it work with the package and the EPC code. But we're almost there!

Well done!!

asf_core_data/getters/mcs/get_mcs_installations.py
ch-williamson: It's here because otherwise lots of the columns end up as objects rather than having consistent types. May be better to specify the types exactly rather than leave it up to this though...

asf_core_data/pipeline/mcs/process/process_mcs_installations.py
ch-williamson: Good idea, done

asf_core_data/pipeline/mcs/process/process_mcs_installations.py
ch-williamson: This was from an email that Alex sent - not sure exactly why it's the case! Will mention this though. DRHI = Domestic RHI - I might look a bit more into the RHI field to see whether it's correct/consistent

asf_core_data/pipeline/mcs/process/process_mcs_installations.py
ch-williamson: Thanks!

asf_core_data/getters/mcs/get_mcs_installations.py
ch-williamson: Great!
This function just loads the data - the saving is done by generate_mcs_data.py and this puts the various versions of processed MCS data in a specific folder (outputs/mcs)

athrado: Oh, my bad. I happened to have a file in /ouputs that fit the format, and I had just assumed it was created by this code. If it's all saved in outputs/mcs then that's perfect!

asf_core_data/pipeline/mcs/process/process_mcs_installations.py
ch-williamson: Agree, this definitely needs more documentation!

Doing a bit more looking into the clusters is on my to-do list - I went with quite a wide interval for the cost estimation tool to make sure we caught as many as possible (based on some EDA) but I'm sure there are some false positives here. That 30% will include new builds too, though I don't really have a comparison to find out whether it's sensible. Comparing properties in clusters is a really good idea - will make sure I do that at some point

asf_core_data/pipeline/mcs/process/process_mcs_installations.py
ch-williamson: Great!

asf_core_data/pipeline/mcs/process/mcs_epc_joining.py
ch-williamson: I think this is doable! I haven't tested this with the most recent EPC data but this merges all EPC columns by default so UPRN should be included too

asf_core_data/pipeline/mcs/process/mcs_epc_joining.py
ch-williamson: Sounds good - yep let's test once everything's on the same branch

asf_core_data/pipeline/mcs/process/mcs_epc_joining.py
ch-williamson: Good idea! Have added

asf_core_data/pipeline/mcs/process/process_mcs_installations.py
ch-williamson: I've added documentation and raised the rest as an issue

asf_core_data/pipeline/mcs/process/mcs_epc_joining.py
ch-williamson: I've made a new issue for making MCS work as a package and have copied over the relevant comments/suggestions

asf_core_data/getters/mcs/get_mcs_installations.py
ch-williamson: I've made a new issue for making MCS work as a package and have copied over the relevant comments/suggestions

asf_core_data/getters/mcs/get_mcs_installations.py
ch-williamson: I've made a new issue for making MCS work as a package and have copied over the relevant comments/suggestions

asf_core_data/getters/mcs/get_mcs_installations.py
ch-williamson: I've made a new issue for making MCS work as a package and have copied over the relevant comments/suggestions

asf_core_data/pipeline/mcs/process/mcs_epc_joining.py
ch-williamson: I think that fillna was added to be cautious about not dropping NAs during deduplication. I might remove it from the loading script (and triple check that nothing gets dropped) and keep it here so that NAs are treated consistently in all fields

ch-williamson: Actually this is really ugly. I'll just remove them here

asf_core_data/pipeline/mcs/process/mcs_epc_joining.py
ch-williamson: Nice idea, done

asf_core_data/getters/mcs/get_mcs_installations.py
ch-williamson: I'll leave this for now as I'd need to apply it to the S3 version anyway, and it's fairly quick, but will make a note to look at this again in the future

asf_core_data/getters/mcs/get_mcs_installations.py
ch-williamson: Will leave for now as this seems to work as expected, but will make a note to look again at dtypes in the future

Issue #9, opened 2022-05-06
Data cleaning revisisted


athrado:
- dtypes
- deal with new values
- better structure for cleaning

PR #15, opened 2022-05-20
EPC Preprocessing and asf-core-data as package

athrado:
Hi Chris, 
thanks for reviewing the code. 

This code review includes many scripts but most of them have been reviewed at an earlier time in context of other projects. You only need to review the following new scripts:

- asf_core_data/getters/epc/data_batches.py
- asf_core_data/utils/visualisation/kepler.py

For these scripts, maybe give it a quick look and check whether the structure makes sense:
- asf_core_data/getters/epc/epc_data.py
- asf_core_data/pipeline/preprocessing/data_cleaning.py

Most scripts only required minor (but tricky) changes (e.g. dtypes, paths, config references etc.) so this review is less about reviewing the code but testing the package. 

I have prepared a Jupyter notebook in /analysis called "ASF Core Data Infrastructure Package testing". So it would be great if you could install the asf-core-data package in another environment and check whether all the code blocks run within that environment. 
Please let me know if there are any dependencies missing or whether it runs into an error at some point.

For this code to run properly, you will need to have a version of the raw EPC data (any version will do for testing) and some of the supplementary data on in local data directory. (I will definitely add functionality so you can directly get it from S3 very soon but for setting up the package I found it easier to work with local data.)

The test code will try to load 2 different batches called 2021_Q2_0721 and 2021_Q4_0721. You can use any EPC raw data batch that you want and rename it like that for testing. You could delete some files from the first batch so you can check whether the number of samples within the different batches really changes. Some of the supplemenary data may need to be downloaded from S3 but shouldn't take long. 

I recommend setting up an ASF data dir and just pass that directory path as LOCAL_DATA_DIR. The ASF data dir should ideally look like this (same structure as on S3). 

Screenshot 2022-05-20 at 11 27 08

Again, this will all be simplier in the future but I just couldn't test the package and the S3 data loads at the same time. I will hopefully also be able to automatically create some output dirs such as /outputs/preprocessed/2021_Q4_0721 - but I'm afraid for now you'll have to manually create that folder.

Thanks a lot for reviewing and please mark anything that looks odd to you or isn't logical. As we're going to re-use that code often, it should be easy and clear to use for all of us. 

If any questions come up (or something ends up too annoying to test the way I suggested), please let me know!

---

This pull request closes the Issues #6, #9, #10, #11 and #12 

Checklist:

- [X] I have refactored my code out from `notebooks/`
- [X] I have checked the code runs
- [X] I have tested the code
- [X] I have run `pre-commit` and addressed any issues not automatically fixed
- [X] I have merged any new changes from `dev`
- [X] I have documented the code
  - [X] Major functions have docstrings
  - [X] Appropriate information has been added to `README`s
- [X] I have explained the feature in this PR or (better) in `output/reports/`
- [X] I have requested a code review


asf_core_data/getters/epc/data_batches.py
```diff
@@ -0,0 +1,139 @@
...
+            Path(obj["Key"]).stem
+            for obj in client.list_objects(Bucket=bucket, Prefix=path, Delimiter="/")[
+                "Contents"
+            ]
+            if obj["Key"].endswith(".zip")
+        ]
+
+    else:
+
+        data_path = get_batch_path(data_path / rel_path.parent, data_path=data_path)
```
ch-williamson, line 86: Maybe move get_batch_path above this one in the script since it's being used here?

asf_core_data/getters/epc/data_batches.py
```diff
@@ -0,0 +1,139 @@
...
+from asf_core_data import Path
+from asf_core_data.config import base_config
+
+import warnings
+import boto3
+
+# ---------------------------------------------------------------------------------
+
+
+def get_all_batch_names(data_path=None, rel_path=base_config.RAW_DATA_PATH):
```
ch-williamson: What's the reasoning behind having a separate data_path and rel_path (and path / data_path below)? Would it make sense to just use a single local_path argument in all of these (which could still take "S3" as a special value)?

asf_core_data/getters/epc/data_batches.py
```diff
@@ -0,0 +1,139 @@
...
+        str: Most recent batch name.
+    """
+
+    if data_path == "S3":
+        batches = get_all_batch_names(data_path="S3")
+
+    else:
+        batches = get_all_batch_names(data_path=data_path, rel_path=rel_path)
+
+    return sorted(batches, reverse=True)[0]
```
ch-williamson: Could be `max(batches)` (but this is fine and clear!)

asf_core_data/getters/epc/data_batches.py
ch-williamson: It doesn't look like these functions work with None as the default path argument (I get a TypeError as it tries to combine None with a Path). Maybe remove the defaults and have these be required?

asf_core_data/getters/epc/data_batches.py
```diff
@@ -0,0 +1,139 @@
...
+        data_path (str/Path, optional): Path to ASF core data directory. Defaults to None.
+        batch (str, optional): Which batch to use, either specific batch name or relative indicator. Defaults to "newest".
+
+    Returns:
+        Path: Path with specific batch name integrated.
+    """
+
+    # If batch does not need to be filled in
+    if not "{}" in str(path):
+        return path
```
ch-williamson: If path is a string, this function returns a string if batch doesn't need to be filled in or a Path otherwise. Is it worth making this `Path(path)` so the output is consistent?

asf_core_data/getters/epc/epc_data.py
```diff
@@ -0,0 +1,683 @@
...
+            dtype=dtype,
+            low_memory=low_memory,
+            usecols=usecols,
+        )
+        for directory in directories
+    ]
+
+    # Concatenate single dataframes into dataframe
+    epc_certs = pd.concat(epc_certs, axis=0)
+    epc_certs["COUNTRY"] = subset
```
ch-williamson, line 118: Would this add a column of Nones in the case where both are loaded?

asf_core_data/getters/epc/epc_data.py
```diff
@@ -0,0 +1,683 @@
...
+            To ensure no mixed types either set False, or specify the type with the dtype parameter.
+
+    Returns:
+        pd.DataFrame: Scotland EPC certificate data for given features."""
+
+    RAW_SCOTLAND_DATA_PATH = data_batches.get_batch_path(
+        Path(data_path) / rel_data_path, data_path=data_path, batch=batch
+    )
+    RAW_SCOTLAND_DATA_ZIP = data_batches.get_batch_path(
+        Path(data_path) / base_config.RAW_ENG_WALES_DATA_ZIP,
```
ch-williamson: Should this be RAW_SCOTLAND_DATA_ZIP?

asf_core_data/getters/epc/epc_data.py
```diff
@@ -0,0 +1,683 @@
...
+    additional_samples = 0
+
+    if subset == "GB" and n_samples is not None:
+        additional_samples = n_samples % 3
+        n_samples = n_samples // 3
+
+    # Get Scotland data
+    if subset in ["Scotland", "GB"]:
+
+        epc_Scotland_df = load_scotland_data(
```
ch-williamson: Better to use all lowercase in the variable name?

asf_core_data/pipeline/preprocessing/data_cleaning.py
```diff
@@ -0,0 +1,515 @@
+# File: asf_core_data/pipeline/preprocessing/data_cleaning.py
+"""Cleaning and standardising the EPC dataset.
+
+
+DATA CLEANING GUIDELINES
```
ch-williamson, line 5: Love these guidelines - super clear!

ch-williamson (commented):
Hello! I've looked at the files you requested and found very little wrong with them :) the one question I had was whether the multiple file path arguments were needed in the batches file, but there's probably a good reason for them! The structure of the epc_data and cleaning files made sense too.

I loaded the package in a notebook and everything went extremely smoothly - no issues with errors or dependencies at all once I'd set up the local file structure correctly!

asf_core_data/getters/epc/data_batches.py
athrado: Didn't realise you could use max for strings. That makes it a bit shorter, thanks!

asf_core_data/getters/epc/data_batches.py
athrado: Good catch! I originally had `.` as default (current dir) but that failed and I guess I changed it to None without testing further.

asf_core_data/getters/epc/epc_data.py
athrado: Good catch, it can't load both at the same time. England and Wales anyway have slightly different columns so I may need to revisit this function first time I need the England recommendations.

asf_core_data/getters/epc/epc_data.py
athrado: it should!!!

asf_core_data/getters/epc/data_batches.py
athrado: It's a) for consistency (since I"m using data_path and rel_path throghout these scripts) and b) for making it easier to work with both S3 and local data. Ideally, S3 and the local data dir are structured the same way so you can always rely on the default rel_path from the config file. So in practice, you never have to set the rel_path.

It's very convenient if you don't have to specify the relative paths and can just pass 'S3' or the local data dir (containing all data) - and the script will know where to look up the specific files.
However, in order to also allow different dir structures, I give the option of passing the relative path. I don't expect this to be used much, though.

The reasoning behind having a data_path and relative_path stems from the problem of no longer having a PROJECT_DIR "for orientation" since it's now a package. I would have loved to just hard-link all paths in the config file but that doesn't work if you have to rely on relative paths as I already knew that we wouldn't want to keep all these large files in the input dir of every single repository. 

Happy to discuss further and test other options - but while it was a bit tricky to set up, I thought it was quite convenient when actually using the package.

asf_core_data/pipeline/preprocessing/data_cleaning.py
athrado: Thanks!

PR #37, opened 2023-02-17
Processing pipeline for historical installers

### LINKED ISSUE #31: Pre-process MCS historical installers data
**Checks prior to pre-processing:**
- [x] Is it possible to merge the historical installers with the installations table, without the certification body?
_Yes, but we need the installer name and the MCS certification number_
- [x] All installers with at least 1 installation have info on the historical installers table?
_No, a small percentage missing._
- [x] Do we have the installer name written in the same way always, independently of the certificate number? Or do we still have duplicates where names are written in slightly different ways?
_We still have duplicates, so we need to create a unique ID to identify the installer. Can be a unique name or a number._
- [x] Only info for installers with any type of HP certifications?
_Yes, although sometimes the start and end dates are in the `_Design Start Date`/`_Design End Date` instead of in the `_Installation Start Date`/`_Installation End Date`. We need to check with MCS what the difference between the Installation and Design variables is._
- [x] `Effective To` > `Effective From`?
_There are 10 instances where `Effective From` is bigger than `Effective to`; 3 instances where both values are the same; Note that `Effective From` can never be missing, but `Effective To` can;_
- [x]  `_End Date` > `_Start Date` ?
_Similar to above (check needs to filter values where `_End Date` is missing)_
- [x] Duplicates (whole instance)?
_1 duplicate_
- [x] For each match when merging the installers to the installations table, is the installation date within Effective From and Effective To?
_Not always._
- [x] Do we have both `_Design_` and `_Installation_` dates?
_Not always, but when we do (ASHP and G/WSHP), values are exactly the same._
- [x] Any variables in the installers snapshot data, not in the historical installers data?
_Yes!_
- [ ] Relationship between `_End Date`/`_Start Date` and `Effective To`/`Effective From`

**To do:**
1. Historical Installers table
- [x] Pre-processing similar to what has previously been done for the MCS installers (snapshot data)
- [x] Deduplicate company name and create unique ID for company
- [x] Create flags for types of certification
- [x] Get certification body (by matching with installations table)
- [x] Add info about installers missing from historical installers table (get info from installations table + Companies House API)
- [x] Create flag to identify records originally in historical installers (to distinguish from those added by us, as point above)
- [x] Create tests for functions
2. Installations table
- [x] Merge with installers to add the newly created unique company ID
3. Fully joined dataset
- [ ] Create a different version that also adds installers info

**Checks after pre-processing:**
- [ ] Companies identified as being the same have the same postcode and address 1?
- [x] Any companies w/ same address 1 and postcode but not identified as being the same? why?

**To think about:**
- [x] Probably better to have a string ID to represent the installer, so that any time new data comes, we make sure the IDs keep the same?

### LINKED ISSUE #45: Change MCS installations processing pipeline to use historical data
Change MCS installations processing pipeline to use historical data instead of concatenating all quarter files. This solves 1 specific problem that was happening:

- [x] since in the last couple of quarters a new variable was added, the values for this variable were appearing as missing data for previous quarters, but the data isn't missing.

sofiapinto:
# Description 

This PR:
- creates a processing pipeline for historical MCS installers data;
- creates getters to retrieve raw historical installers and installations data from S3 (I've added the data manually to S3, to `asf-core-data/inputs/MCS/latest_raw_data/historical/` - had to create a folder for historical data within `latest_raw_data`, otherwise it would have messed the work in `generate_mcs_data.py` because of the file names);
- adds the company unique ID to the installations data in the processing installations pipeline;
- creates tests for the processing pipeline;
- I've also added a notebook (`asf_core_data/analysis/historical_installers.ipynb`) if you want to take a look at raw data and processed data, while reviewing the code, to better understand the changes.

closes #31 
closes #45

Hey @ch-williamson, @athrado and @sqr00t thanks so much for taking the time to look at this PR.
- Chris and Julia, it would be great if at least one of you could check the files run, but more than that, check if all makes sense within the asf_core_data data processing pipeline and that the additions/changes don't have potential for messing something else up :) as a follow-up we should probably add a few tests that are automatically run by github actions.
- Solomon, it would be great if you could check if the general "flow" looks good to you/makes sense and if any changes are needed before the code gets transferred to asf_daps. Note that, the historical installers data needs to have a pair of variables as table primary key: `(company_unique_id, mcs_certificate_number)`
- Julia, I didn't add code to create a merged version of the data (for HPMT consumption) that includes the installers, as I didn't have time to familiarise enough with `mcs_epc_joining.py` and `merge_install_dates.py` scripts, to integrate these changes somewhere in the code. But hopefully it shouldn't be too difficult. All that needs to be done is an outer merge between processed installers and installations. Happy to do that after I'm back, or for you to do it if you would prefer.

## Setup
- Create a developer account for Companies House API (it's super quick to set up, but the confirmation email might take a few minutes to arrive, maybe about 10-15min):
   1. Register for an account [here](https://developer.company-information.service.gov.uk/get-started);
   2. Create a new application [here](https://developer.company-information.service.gov.uk/manage-applications/add). Give it any name and description you want and choose the Live environment.
   3.  Copy your API key credentials.
- checkout to the correct branch: `git checkout 31_historical_MCS_installers_data`
- Run `make install`;
- Run `direnv allow`;

## Review
Could you run the following script to check if it works for you?
`python3 asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py`
A file will be saved to S3 and replace the one I have previously created (absolutely no problem with that, feel free to run!). It gives a warning, `SettingWithCopyWarning`, but all should run fine and a file should be saved to S3: `asf-core-data/outputs/MCS/installers/mcs_historical_installers_20230207.csv`

There's a script to test the functions I created. Could you please check if 17 out 18 tests pass? Should take approximately 2 mins to run.
`pytest asf_core_data/pipeline/mcs/test/test_process_historical_mcs_installers.py`
One of the tests fails, the one testing the data coming from Companies House - it doesn't mean that the function is not doing what's supposed to do. It just means that 1) the accuracy of the matching is low or 2) the matching is good but the data is incorrect in one of the sources (when comparing postcode data coming from MCS to Companies House data). I've manually checked some of the values and it seems that sometimes Companies House has the correct postcode, other times the correct postcode is in MCS data. But I need to take a further look at this!
Also please feel free to suggest other tests.

I would also ask you to review the remaining scripts with changes, with a bigger focus on: 
- `asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py`
- The additions made to `asf_core_data/pipeline/mcs/process/process_mcs_installations.py` - is this a good place to add the "company_unique_id" var to the installations? Or maybe it should be added to `generate_mcs_data.py`?

A few questions:
- Shall we keep the postcode with the space or without?
- Does the general pre-processing pipeline for historical installers make sense? Any suggestions for simplification/to improve readability?

Thanks a lot!

---

# Checklist:

- [x] I have refactored my code out from `notebooks/`
- [x] I have tested the code
  - [x] Major functions have docstrings

### DIFF
--- a/asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
+++ b/asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
@@ -1,5 +1,5 @@
 """
-Script for pre-processing historical MCS heat pump installer company data:
+Script for pre-processing historical MCS heat pump installer company data.
 - Dropping duplicate instances (if they exist);
 - Renaming columns;
 - Joining installation and design variables;
@@ -29,11 +29,6 @@ OR
 `python3 process_historical_mcs_installers.py -raw_historical_installers_filename "raw_historical_mcs_installers_YYYYMMDD.xlsx" -raw_historical_installations_filename "raw_historical_mcs_installations_YYYYMMDD.xlsx"`
 OR
 `python3 process_historical_mcs_installers.py -api_key "YOUR_API_KEY" -raw_historical_installers_filename "raw_historical_mcs_installers_YYYYMMDD.xlsx" -raw_historical_installations_filename "raw_historical_mcs_installations_YYYYMMDD.xlsx"`
-
-Companies House API additional info:
-- https://developer-specs.company-information.service.gov.uk/companies-house-public-data-api/reference/search/search-companies
-- https://developer-specs.company-information.service.gov.uk/companies-house-public-data-api/resources/companysearch?v=latest
-
 """
 
 import pandas as pd
@@ -44,7 +39,7 @@ from argparse import ArgumentParser
 from asf_core_data.config import base_config
 from asf_core_data.getters.data_getters import s3, load_s3_data, save_to_s3
 from asf_core_data.getters.mcs_getters.get_mcs_installers import (
-    get_raw_historical_mcs_installers,
+    get_raw_historical_installers_data,
 )
 from asf_core_data.getters.mcs_getters.get_mcs_installations import (
     get_raw_historical_installations_data,
@@ -674,15 +669,11 @@ if __name__ == "__main__":
     uk_geo_data = load_s3_data(s3_bucket_name, uk_geo_path)
 
     # Getting raw installations and installers data from S3
-    mcs_raw_data_path = base_config.MCS_HISTORICAL_DATA_INPUTS_PATH
-    raw_historical_installers_filename = args.raw_historical_installers_filename
-    raw_historical_installations_filename = args.raw_historical_installations_filename
-
-    raw_historical_installers = get_raw_historical_mcs_installers(
-        raw_historical_installers_filename
+    raw_historical_installers = get_raw_historical_installers_data(
+        args.raw_historical_installers_filename
     )
     raw_historical_installations = get_raw_historical_installations_data(
-        raw_historical_installations_filename
+        args.raw_historical_installations_filename
     )
 
     # Process raw historical installers data
@@ -695,12 +686,10 @@ if __name__ == "__main__":
 
     # Saving processed data to S3
     processed_data_path = base_config.PREPROCESSED_MCS_HISTORICAL_INSTALLERS_FILE_PATH
-
+    # date when data was shared by MCS in MCS data dumps
     date = args.raw_historical_installers_filename.split("installers_")[1].split(
         ".xlsx"
-    )[
-        0
-    ]  # date when data was shared by MCS in MCS data dumps
+    )[0]
 
     save_to_s3(
         s3=s3,
--- a/asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
+++ b/asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
@@ -190,32 +190,12 @@ def match_companies_house(company_name: str, api_key: str) -> pd.Series:
             and ("items" in company_data.keys())
             and len(company_data["items"]) > 0
         ):
-            # return data for the closest match, the one found in index 0
-            if ("address_snippet" in company_data["items"][0].keys()) and (
-                "date_of_creation" in company_data["items"][0].keys()
-            ):
-                return pd.Series(
-                    [
-                        company_data["items"][0]["address_snippet"],
-                        company_data["items"][0]["date_of_creation"],
-                    ]
-                )
-            elif "address_snippet" in company_data["items"][0].keys():
-                return pd.Series(
-                    [
-                        company_data["items"][0]["address_snippet"],
-                        None,
-                    ]
-                )
-            elif "date_of_creation" in company_data["items"][0].keys():
-                return pd.Series(
-                    [
-                        None,
-                        company_data["items"][0]["date_of_creation"],
-                    ]
-                )
-            else:
-                return pd.Series([None, None])
+           address_snippet = company_data["items"][0]["address_snippet"] if ("address_snippet" in company_data["items"][0].keys()) else None
+           date_of_creation = company_data["items"][0]["date_of_creation"] if  "date_of_creation" in company_data["items"][0].keys() else None
+           
+           return  pd.Series(
+                    [address_snippet, date_of_creation])
+        
         else:
             return pd.Series([None, None])
     else:
--- a/asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
+++ b/asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
@@ -411,7 +411,7 @@ def create_certified_flags(data: pd.DataFrame):
     Updates historical installers data by creating flags representing whether
     installer is certified or not for a certain technology.
 
-    If a start date exists, then if means installer is certified.
+    If a start date exists, then it means installer is certified.
 
     Args:
         data: historical installers data
--- a/asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
+++ b/asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
@@ -3,11 +3,11 @@ Script for pre-processing historical MCS heat pump installer company data:
 - Dropping duplicate instances (if they exist);
 - Renaming columns;
 - Joining installation and design variables;
+- Adding certification body information.
 - Adding instances for installers with installations but with info missing from installers table;
 - Creating flag variables for whether an installer is certified for a certain technology;
 - Geocoding data by mapping postcode to latitude and longitude values;
 - Creating a variable that uniquely identifies installers;
-- Adding certification body information.
 
 Note that:
 - installer name/ company name/ installation company name are used interchangebly throughout the whole script;
@@ -38,9 +38,6 @@ Companies House API additional info:
 
 import pandas as pd
 import numpy as np
-import requests
-import random
-import time
 import os
 from datetime import datetime
 from argparse import ArgumentParser
@@ -51,6 +48,16 @@ from asf_core_data.getters.mcs_getters.get_mcs_installers import (
 )
 from asf_core_data.getters.mcs_getters.get_mcs_installations import (
     get_raw_historical_installations_data,
+)
+from asf_core_data.pipeline.mcs.process.process_mcs_utils import (
+    rename_columns,
+    match_companies_house,
+    clean_company_name,
+    geocode_postcode,
+    deal_with_versions_of_trading_as,
+    from_list_to_dictionary,
+    map_position_of_subset_items,
+    position_to_value,
 )
 
 
@@ -78,27 +85,6 @@ def basic_preprocessing_of_installations(raw_historical_installations: pd.DataFr
     ] = raw_historical_installations["installation_company_mcs_number"].astype(int)
 
 
-def rename_columns(cols: list) -> list:
-    """
-    Renames a list of strings, e.g. a list of column names, by:
-    - applying lower case;
-    - replacing "heat pump" by "hp";
-    - replacing " " and "/" by "_";
-    - correcting the word address.
-
-    Args:
-        cols: original column names
-    Returns:
-        Renamed column names
-    """
-    cols = [c.lower() for c in cols]
-    changes_dict = {"heat pump": "hp", " ": "_", "/": "_", "adddress": "address"}
-    for item in changes_dict.keys():
-        cols = [c.replace(item, changes_dict[item]) for c in cols]
-
-    return cols
-
-
 def join_installation_and_design_vars(data: pd.DataFrame):
     """
     Joins together installation and design date variables and drops original variables.
@@ -124,99 +110,6 @@ def join_installation_and_design_vars(data: pd.DataFrame):
         )
 
         data.drop([col, correspondent_design_var_name], axis=1, inplace=True)
-
-
-def deal_with_versions_of_trading_as(installer_name: str) -> str:
-    """
-    Replaces several expressions representing "trading as" by the expression
-    "trading as" in the installer name.
-
-    Args:
-        installer_name: installer name
-
-    Returns:
-        updated installer name
-    """
-    versions = {
-        "t/a": "trading as",
-        " ta ": " trading as ",
-        "a trading name of": "trading as",
-        "the trading name of": "trading as",
-    }
-    for item in versions:
-        installer_name = installer_name.replace(item, versions[item])
-
-    return installer_name
-
-
-# copied from process_mcs_utils.py and changed slightly
-def match_companies_house(company_name: str, api_key: str) -> pd.Series:
-    """
-    Fuzzy match between MCS company name and Companies House API company names and
-    returns Companies House information: company full address and company date of creation.
-
-    Args:
-        company_name: Company name used to query company house API
-        api_key: API key to request data from live company house API endpoint
-    Returns:
-        company_info: pandas Series with 2 values, full address and creation date
-    """
-
-    # E.g. "Flo Group LTD T/A Flo Renewables" -> "flo renewables"
-    company_name = company_name.lower()
-    if "trading as" in company_name:
-        company_name = deal_with_versions_of_trading_as(company_name).split(
-            " trading as "
-        )[1]
-
-    # endpoint url
-    base_url = f"https://api.company-information.service.gov.uk/search/companies?q={company_name}"
-
-    response = requests.get(base_url, auth=(api_key, ""))
-
-    # developer guidelines state you can make 600 requests per 5 mins
-    # this translates into 1 request every 0.5 seconds
-    time.sleep(0.5)
-
-    response_status_code = response.status_code
-
-    if response_status_code == 200:  # status code 200 means all good with request
-        company_data = response.json()
-        if (
-            (company_data is not None)
-            and ("items" in company_data.keys())
-            and len(company_data["items"]) > 0
-        ):
-            address_snippet = (
-                company_data["items"][0]["address_snippet"]
-                if ("address_snippet" in company_data["items"][0].keys())
-                else None
-            )
-            date_of_creation = (
-                company_data["items"][0]["date_of_creation"]
-                if "date_of_creation" in company_data["items"][0].keys()
-                else None
-            )
-
-            return pd.Series([address_snippet, date_of_creation])
-
-        else:
-            return pd.Series([None, None])
-    else:
-        if response_status_code >= 400 and response_status_code < 500:
-            raise Exception(
-                "Cannot get data, the program will stop!\nHTTP {}: {}".format(
-                    response_status_code, response.text
-                )
-            )
-        sleep_seconds = random.randint(5, 60)
-        print(
-            "Cannot get data, your program will sleep for {} seconds...\nHTTP {}: {}".format(
-                sleep_seconds, response_status_code, response.text
-            )
-        )
-        time.sleep(sleep_seconds)
-        return match_companies_house(company_name, api_key)
 
 
 def recompute_full_adress(
@@ -429,42 +322,6 @@ def create_certified_flags(data: pd.DataFrame):
         data[flag_var] = ~pd.isnull(data[col])
 
 
-# copied from process_mcs_utils.py and changed slightly
-def clean_company_name(company_name: str) -> str:
-    """
-    Cleans installation company name by:
-        - changing company name to lower case;
-        - removing punctuation;
-        - removing company-related stopwords like "ltd", "limited", etc.;
-
-    Args:
-        company_name: Name of company to clean.
-    Returns:
-        company_name: Cleaned company name.
-    """
-    company_stopwords = ["ltd", "limited", "old", "account"]
-
-    company_name = company_name.lower().translate(str.maketrans("", "", "()-,/.&"))
-    company_name = [
-        comp for comp in company_name.split() if comp not in company_stopwords
-    ]
-
-    return " ".join(company_name)
-
-
-def from_list_to_dictionary(list_values: list) -> dict:
-    """
-    Transforms a list into a dictionary where each key is matched to the first
-    value in the list. Creates a mapping between each item in the list and the first one.
-    Arg:
-        list_values: a list of values e.g. ["Company A", "Company B", "Company A trading as Company B"]
-    Returns:
-        Dictionary mapping all values to the first one
-        e.g. {"Company B":"Company A", "Company A trading as Company B": "Company A"}
-    """
-    return {value: list_values[0] for value in list_values[1:]}
-
-
 def dictionary_mapping_trading_as_company_names(
     trading_as_companies: pd.DataFrame,
 ) -> dict:
@@ -509,44 +366,6 @@ def dictionary_mapping_trading_as_company_names(
     }
 
     return trading_as_dictionary
-
-
-def map_position_of_subset_items(list_of_strings: list) -> dict:
-    """
-    Returns a mapping between the position of strings in a list which are a subset of each other.
-    Example:
-    set_of_strings = ["company a ltd", "a ltd", "a"]
-    Outputs {0: 2, 1: 2}
-    Note that, in the example above, 0 is initially mapped to 1 but then that's overwritten by 0 being mapped to 2.
-
-    Args:
-        list_of_strings: a list of strings
-    Returns:
-        A dictionary mapping the position of strings that are subset of another.
-    """
-    return {
-        i: j
-        for i in range(len(list_of_strings) - 1)
-        for j in range(i + 1, len(list_of_strings), 1)
-        if (list_of_strings[i] in list_of_strings[j])
-        or (list_of_strings[j] in list_of_strings[i])
-    }
-
-
-def position_to_value(list_of_strings: list, indices_dict: dict) -> dict:
-    """
-    Returns dictionary resulting from mapping a dictionary with indices as key-value pairs to a list of strings.
-    Args:
-        list_of_strings: list of strings, e.g. ["a", "b", "c"]
-        indices_dict: dictionary of indices, e.g. {0:2, 1:2}
-    Returns:
-        The mapped dictionary e.g. {"a":"c", "b":"c"}
-
-    """
-    return {
-        list_of_strings[key]: list_of_strings[indices_dict[key]]
-        for key in indices_dict.keys()
-    }
 
 
 def map_installers_same_location_different_id(
@@ -673,28 +492,6 @@ def create_installer_unique_id(installer_data: pd.DataFrame):
     )
 
 
-# copied from process_mcs_utils.py and changed slightly
-def geocode_postcode(data: pd.DataFrame, geodata: pd.DataFrame) -> pd.DataFrame:
-    """
-    Updates data with latitude and longitude columns, by merging with geodata
-    on postode column (left merge, not to loose any data).
-    Also transforms postcode column by removing the space.
-
-    Args:
-        data: DataFrame with postcode column.
-        geodata: DataFrame with postcode, latitude and longitude columns.
-    """
-
-    geodata["postcode"] = geodata["postcode"].str.replace(" ", "")
-
-    data["postcode"] = data["postcode"].str.upper().str.replace(" ", "")
-    data = data.merge(
-        geodata[["postcode", "latitude", "longitude"]], how="left", on="postcode"
-    )
-
-    return data
-
-
 def add_certification_body_info(
     installer_data: pd.DataFrame, installations_data: pd.DataFrame
 ) -> pd.DataFrame:
@@ -741,9 +538,6 @@ def add_certification_body_info(
     # Dropping installation variables used to merge and match
     installer_data.drop(columns=installations_match_vars, inplace=True)
 
-    print("installer columns check:")
-    print(installer_data.columns)
-
     return installer_data
 
 
@@ -758,11 +552,11 @@ def preprocess_historical_installers(
     - Dropping duplicate instances (if they exist);
     - Renaming columns;
     - Joining installation and design variables;
+    - Adding certification body information;
     - Adding instances for installers with installations but with info missing from installers table;
     - Creating flag variables for whether an installer is certified for a certain technology;
     - Geocoding data by mapping postcode to latitude and longitude values;
-    - Creating a variable that uniquely identifies installers;
-    - Adding certification body information.
+    - Creating a variable that uniquely identifies installers.
 
     Args:
         raw_historical_installers: raw historical installers data
--- a/asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
+++ b/asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
@@ -395,16 +395,13 @@ def update_full_address(
         full adress
     """
     if original_record:
-        temp = address_1
-        if not pd.isnull(address_2):
-            temp = temp + address_2 + ", "
-        if not pd.isnull(town):
-            temp = temp + town + ", "
-        if not pd.isnull(county):
-            temp = temp + county + ", "
-        if not pd.isnull(postcode):
-            temp = temp + postcode + ", "
-        return temp[:-2]
+        temp = []
+        for address_part in [address_1, address_2, town, county, postcode]:
+              if not pd.isnull(address_part):
+                    temp.append(address_part)
+        full_address = ", ".join(temp)
+        return full_address
+ 
     # if not original record, we have the full address from Companies House
     return full_address
 
--- a/asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
+++ b/asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
@@ -29,6 +29,9 @@ OR
 `python3 process_historical_mcs_installers.py -raw_historical_installers_filename "raw_historical_mcs_installers_YYYYMMDD.xlsx" -raw_historical_installations_filename "raw_historical_mcs_installations_YYYYMMDD.xlsx"`
 OR
 `python3 process_historical_mcs_installers.py -api_key "YOUR_API_KEY" -raw_historical_installers_filename "raw_historical_mcs_installers_YYYYMMDD.xlsx" -raw_historical_installations_filename "raw_historical_mcs_installations_YYYYMMDD.xlsx"`
+
+Note that the processing of historical installers is included the script that processes MCS data:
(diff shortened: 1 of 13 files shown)


asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+        .str.split(" ")
+        .str[0]
+    )
+
+    raw_historical_installations["installation_company_mcs_number"] = (
+        raw_historical_installations["installation_company_mcs_number"]
+        .str.split(" ")
+        .str[1]
+        .astype(int)
+    )
```
athrado: All good. In case you want to save a few lines, you could do this in one step. Look at dummy example below. You'd still have to modify the company_mcs_number with `.astype(int)`, though.

Screenshot 2023-02-25 at 15 02 17

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+    - replacing " " and "/" by "_";
+    - correcting the word address.
+
+    Args:
+        cols: original column names
+    Returns:
+        Renamed column names
+    """
+    cols = [c.lower() for c in cols]
+    changes_dict = {"heat pump": "hp", " ": "_", "/": "_", "adddress": "address"}
```
athrado: Didn't check whether this ends up being a problem in practice, but since dicts are not ordered, I worry that the following could happen: In the string 'heat pump', the white space is replaced with '_' and so 'heat_pump' will not be shortened to 'hp'. Easiest fix is adding an additonal replacement entry:

```suggestion
    changes_dict = {"heat pump": "hp", "heat_pump":"hp", " ": "_", "/": "_", "adddress": "address"}
```

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+
+    installation_cols = [col for col in data.columns if "installation" in col]
+    for col in installation_cols:
+        new_date_var = col.replace("installation_", "")
+        correspondent_design_var_name = col.replace("installation", "design")
+        data[new_date_var] = data.apply(
+            lambda x: x[col]
+            if not pd.isnull(x[col])
+            else x[correspondent_design_var_name],
+            axis=1,
```
athrado, line 107: Looks fine and I assume you tested that this does exactly what it should, but I haven't double-checked myself.

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+        updated installer name
+    """
+    versions = {
+        "t/a": "trading as",
+        " ta ": " trading as ",
+        "a trading name of": "trading as",
+        "the trading name of": "trading as",
+    }
+    for item in versions:
+        installer_name = installer_name.replace(item, versions[item])
```
athrado: Looks good! Would it be worth adding a `.strip()` at the end of the installer name to get rid of any trailing while spaces, or does that never happen?

athrado: Looking at the code further down, I guess that's not a problem.

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+
+    return installer_name
+
+
+# copied from process_mcs_utils.py and changed slightly
+def match_companies_house(company_name: str, api_key: str) -> pd.Series:
+    """
```
athrado: Nicely done and well-explained!

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+                )
+            elif "date_of_creation" in company_data["items"][0].keys():
+                return pd.Series(
+                    [
+                        None,
+                        company_data["items"][0]["date_of_creation"],
+                    ]
+                )
+            else:
+                return pd.Series([None, None])
```
athrado: If I understand correctly, this could be handled with less code/make it a bit more reabable. Maybe something like:

```suggestion
           address_snippet = company_data["items"][0]["address_snippet"] if ("address_snippet" in company_data["items"][0].keys()) else None
           date_of_creation = company_data["items"][0]["date_of_creation"] if  "date_of_creation" in company_data["items"][0].keys() else None
           
           return  pd.Series(
                    [address_snippet, date_of_creation])
        
```

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+        if response_status_code >= 400 and response_status_code < 500:
+            raise Exception(
+                "Cannot get data, the program will stop!\nHTTP {}: {}".format(
+                    response_status_code, response.text
+                )
+            )
+        sleep_seconds = random.randint(5, 60)
+        print(
+            "Cannot get data, your program will sleep for {} seconds...\nHTTP {}: {}".format(
+                sleep_seconds, response_status_code, response.text
```
athrado: Very nice way of handling this! It actually happened to me so I know it's working!

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+    Args:
+        date_creation: date company was created, according to match from Companies House
+        first_comissioning_date: first HP comissioning date according to HP installations data
+        full_address: full address retrieved from Companies House match
+    Returns:
+        recomputed full address
+    """
+    if pd.isnull(date_creation) or pd.isnull(first_commisioning_date):
+        return full_address
+    return full_address if date_creation <= first_commisioning_date else None
```
athrado, line 130: Makes sense!

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+        axis=1,
+        inplace=True,
+    )
+
+    # dropping unecessary variables
+    missing_installers.drop(
+        ["date_of_creation"],
+        axis=1,
+        inplace=True,
+    )
```
athrado: For .rename and .drop functions, you can also use they keyword argument 'columns' instead of defining the axis. Doens't really matter but I'm always paranoid about mixing up the axes. :D 

```suggestion
    missing_installers.rename(columns=
        {
            "installation_company_name": "company_name",
            "installation_company_mcs_number": "mcs_certificate_number",
        }
        inplace=True,
    )

    # dropping unecessary variables
    missing_installers.drop(columns=
        ["date_of_creation"],
        inplace=True,
    )
```

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+        temp = address_1
+        if not pd.isnull(address_2):
+            temp = temp + address_2 + ", "
+        if not pd.isnull(town):
+            temp = temp + town + ", "
+        if not pd.isnull(county):
+            temp = temp + county + ", "
+        if not pd.isnull(postcode):
+            temp = temp + postcode + ", "
+        return temp[:-2]
```
athrado: That looks fine, but you could also try something like to prevent repetition and confusion about `temp[:-2]`

```suggestion
    if original_record:
        temp = []
        for address_part in [address_1, address_2, town, county, postcode]:
              if not pd.isnull(address_part):
                    temp.append(address_part)
        full_address = ", ".join(temp)
        return full_address
 
```

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+    # if not original record, we have the full address from Companies House
+    return full_address
+
+
+def create_certified_flags(data: pd.DataFrame):
+    """
+    Updates historical installers data by creating flags representing whether
+    installer is certified or not for a certain technology.
+
+    If a start date exists, then if means installer is certified.
```
athrado:
```suggestion
    If a start date exists, then it means installer is certified.
```

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+        company_name: Cleaned company name.
+    """
+    company_stopwords = ["ltd", "limited", "old", "account"]
+
+    company_name = company_name.lower().translate(str.maketrans("", "", "()-,/.&"))
+    company_name = [
+        comp for comp in company_name.split() if comp not in company_stopwords
+    ]
+
+    return " ".join(company_name)
```
athrado: I assume it's fine that the result is lowercase.

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+    Arg:
+        list_values: a list of values e.g. ["Company A", "Company B", "Company A trading as Company B"]
+    Returns:
+        Dictionary mapping all values to the first one
+        e.g. {"Company B":"Company A", "Company A trading as Company B": "Company A"}
+    """
+    dictionary = dict()
+    for value in list_values[1:]:
+        dictionary[value] = list_values[0]
+    return dictionary
```
athrado: Why would we need this...? I'll probably find out later but it seems odd.  

Either way, you can do this much simpler (see also dummy example). Hardly worth having it's own function. 

Screenshot 2023-02-25 at 17 04 23

```suggestion
def from_list_to_dictionary(list_values):
    """
    Transforms a list into a dictionary where each key is matched to the first
    value in the list. Creates a mapping between each item in the list and the first one.
    Arg:
        list_values: a list of values e.g. ["Company A", "Company B", "Company A trading as Company B"]
    Returns:
        Dictionary mapping all values to the first one
        e.g. {"Company B":"Company A", "Company A trading as Company B": "Company A"}
    """
    return {value : list_values[0] for value in list_values[1:]}
```

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+    They should all be identified as the same company, e.g. identified as "Company A".
+
+    Args:
+        trading_as_companies: dataframe containing a processed_company_name variable where
+        company names contain the "trading as" expression.
+    Returns:
+        A dictionary mapping company names.
+    """
+
+    # For each instance in data, it creates a ist with all possible versions of each company names
```
athrado:
```suggestion
    # For each instance in data, it creates a list with all possible versions of each company names
```

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+    # splitting string by "trading as". E.g. "A trading as B" -> ["A", "B"]
+    trading_as_companies["list"] = trading_as_companies[
```
athrado: I think `list` and `dictionary` are not suitable column names. More specific?

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+    # Adding full name to the list as well : ["A", "B", "A trading as B"]
+    trading_as_companies["list"] = trading_as_companies.apply(
+        lambda x: x["list"] + [x["processed_company_name"]], axis=1
+    )
+
+    # Transforms the list into a dictionary where each key is matched to the first value in the list:
+    # ["A", "B", "A trading as B"] -> {"B":"A", "A trading as B":"A"}
+    trading_as_companies["dictionary"] = trading_as_companies["list"].apply(
+        from_list_to_dictionary
+    )
```
athrado: Is there an advantage of saving the dicts in the column 'dictionary' instead of just concatenating all dicts right away...? I may also misunderstand what's happening here but this could potentially be simplified.

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+    new_matches = map_installers_same_location_different_id(
+        same_postcode_no_match, installer_data
+    )
+    installer_data["company_unique_id"] = installer_data["company_unique_id"].apply(
+        lambda x: new_matches[x] if x in new_matches.keys() else x
+    )
+
+
+# copied from process_mcs_utils.py and changed slightly
+def geocode_postcode(data: pd.DataFrame, geodata: pd.DataFrame) -> pd.DataFrame:
```
athrado: If we don't need the coordinates for this script itself, I would recommend we add the postcode (and all the other geographical data) at the very end after merging (EPC, MCS installations, MCS installers) to keep it simple. Otherwise we'll have code for extracting the long/lats from post in every script.

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+        raw_historical_installers.columns
+    )
+
+    # Joining installation and design variables
+    join_installation_and_design_vars(raw_historical_installers)
+
+    # flag "original_records" (if originally in installers or inputed from installations)
+    raw_historical_installers["original_record"] = [
+        True for i in range(len(raw_historical_installers))
+    ]
```
athrado, line 585: I assume that works, but make sure the order is correct. Are the first aways the original records?

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+    raw_historical_installers = add_certification_body_info(
+        raw_historical_installers, raw_historical_installations
+    )
+
+    return raw_historical_installers[
+        base_config.processed_historical_installers_columns_order
+    ]
+
+
+def create_argparser():
```
athrado: This is something that should probably go to utils

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+
+if __name__ == "__main__":
+    s3_bucket_name = base_config.BUCKET_NAME
+
+    parser = create_argparser()
+    args = parser.parse_args()
+
+    # Loading geodata
+    uk_geo_path = base_config.POSTCODE_TO_COORD_PATH
+    uk_geo_data = load_s3_data(s3_bucket_name, uk_geo_path)
```
athrado, line 672: Again, this might be premature. I would not worry about the geodata yet.

athrado (commented):
Hey Sofia,

_For now I've only checked the script `process_historical_mcs_installers.py` - but I'll have another go at the rest later._ 

It's super impressive how you handle all the processing steps and clean up and enhance the data. I can tell that a lot of work went into the details of the code and you documented everything nicely. It's clear that you know exactly what you're doing and why it needs to be done. Well done!

I haven't found any major issues with the code, but I have left a few suggestions for how to improve the code, make it more readable or simplify it. Maybe some of it is helpful, but feel free to discard if it doesn't seem suitable to you.

If there is time, I would recommend splitting up this script into multiple scripts with one main pipeline script that uses code from other modules. I think there are a few functions that logically form a block, so they could go together into one module. I think that would make the script more digestible. (I'm myself guilty of having very long scripts, so I totally understand how easily this happens.) 

And while I'm fine with it, I think that pandas-critical people would be very skeptical of all the pandas operations in your code, especially the complex `.apply()` with lambda expressions. I also don't have a better/fast/cleaner solution for that at hand and it's well-documented so fine by me, but I felt I should point it out anyways... Maybe we can learn about alternatives together at some point. ;-) 

I ran `python3 asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py` and it ran through without errors after fixing the `certification_body` colum at the end (see comment in the code review).  The pytest script also yielded the expected 17 out of 18. 

_I'll have another look at the rest next week but from what I've seen there's not much to correct anymore._

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,926 @@
...
+    ]
+
+    # Merge installer with installation data to add the certification_body to the installer data
+    installer_data = installer_data.merge(
+        unique_pairs_in_installations,
+        how="left",
+        right_on=installations_match_vars,
+        left_on=installers_match_vars,
+    )
+
```
sqr00t, line 535: When this function is called, the input installers and installations dataframes already have certification_body. Merge results in certification_body with _x and _y suffixes. Suggestion to assign _left and _right suffixes respectively for clarity, then merge _right data into _left if there are nulls in _left, using combine_first.

```
    # Merge installer with installation data to add the certification_body to the installer data
    installer_data = installer_data.merge(
        unique_pairs_in_installations,
        how="left",
        left_on=installers_match_vars,
        right_on=installations_match_vars,
        suffixes=('_left','_right')
    )

    # Unify _left and _right certification_body, filling any left nulls with right values
    installer_data["certification_body"] = \
        installer_data["certification_body_left"].combine_first(
            installer_data["certification_body_right"]
        )

File: asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py [hunk 22]
```diff
@@ -0,0 +1,926 @@
...
+    raw_historical_installers = geocode_postcode(
+        raw_historical_installers, geographical_data
+    )
+
+    # Create installer unique ID
+    create_installer_unique_id(raw_historical_installers)
+
+    # Certification body information
+    raw_historical_installers = add_certification_body_info(
+        raw_historical_installers, raw_historical_installations
```
sqr00t: Line 793 adds 'certification_body' to installations dataframe. Lines 812 to 818 adds 'certification_body' to installers dataframe. Resulting in duplicate 'certification_body' in merge. Please see suggestion referred in this function's definition (Lines 752 to 759) to account for this.

sqr00t (changes requested):
Suggestion for duplicate certification_body resulting from add_certification_body_info function

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: definitely! thank you :D

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: cool :) really glad

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: great! :D

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: very neat! will add :)

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: thank you :)

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: yep, it's fine, because we don't do these changes to the original company name. We apply them to get the company unique id :)

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: At this point in the code we only have original records. Only after this, lines 806-809, we add the missing installers, i.e.  non-original records.

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: A few other functions should definitely go to a utils file. But I don't think this one should go, as it is only going to be used in this file and can't be used anywhere else. Am I seeing wrong? :) @athrado

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: yep, using `columns` or `axis=1` is the same, but i'll change it :) it's definitely easier to understand with `columns` :)

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: I don't think this would ever be a problem in practice, but I will add a test to check this in the test script :)

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: Yup, I have a test to check the output from this function in `test_process_mcs_historical_installers.py`, the function `test_join_installation_and_design_vars()` - but actually think I just found a bug in the test!

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: Nop, I think the `clean_company_name()` function takes care of that :)
I can add a further test to the `test_process_historical_mcs_installers.py` script, but all trailing white spaces are dealt with in that function.

Example:
```
company_name = "\tabc t/a  b"
company_name = clean_company_name(company_name)
```
The value of company_name is 'abc ta b'

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: very cool suggestion, didn't know about `expand`. Will change, thanks so much! :D

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: Yup, I can change those to `list_of_companies` and `dictionary_of_companies`, respectively.

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: Saving to the column dictionary is just an intermediate step, but happy to do it all in one go. Any suggestions? @athrado

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: I'm adding the coordinates as this was also previously done by India in the work she did on the installers snapshot data, and it might be useful for a few analyses. Just trying to make this as broad as possible, to account for other analyses besides HPMT. :)

asf_core_data/pipeline/mcs/process/process_mcs_installations.py
```diff
@@ -117,7 +120,22 @@ def identify_clusters(hps, time_interval=base_config.MCS_CLUSTER_TIME_INTERVAL):
     return hps
 
 
-#####
+def get_installer_unique_id(installations, installers):
+    """
+    Updates installations table by adding the unique installer ID.
+    Args:
+        installations: installations table
+        installers: historical installers table
```
athrado, line 132: maybe good to mark what dtype is expected, either in doc string or type hint

asf_core_data/pipeline/mcs/process/process_mcs_installations.py
```diff
@@ -133,11 +151,16 @@ def get_processed_installations_data(refresh=True):
...
+
     installations_data = add_hp_features(installations_data)
     installations_data = mask_outliers(installations_data)
     installations_data = identify_clusters(installations_data)
-    # installations_data["installer_name"] = installations_data["installer_name"].apply(
-    #     clean_company_name
-    # )
+
+    # Adding variable with unique installer ID
+    get_installer_unique_id(installations_data, historical_installers_processed_data)
```
athrado: `get_installer_unique_id()` doesn't return anything...? 
Where does the data with added unique ID go? Sorry if I missed something.

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,905 @@
...
+        print(
+            "Cannot get data, your program will sleep for {} seconds...\nHTTP {}: {}".format(
+                sleep_seconds, response_status_code, response.text
+            )
+        )
+        time.sleep(sleep_seconds)
+        return match_companies_house(company_name, api_key)
+
+
+def recompute_full_adress(date_creation, first_commisioning_date, full_address):
```
athrado: Type hint or doc string

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
athrado: True, maybe once the rest is more modularised this can stay here. It just struck me as 'out-of-character' for this script

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,905 @@
...
+
+    company_name = company_name.lower().translate(str.maketrans("", "", "()-,/.&"))
+    company_name = [
+        comp for comp in company_name.split() if comp not in company_stopwords
+    ]
+
+    return " ".join(company_name)
+
+
+def from_list_to_dictionary(list_values):
```
athrado: type hint for consistency

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -0,0 +1,905 @@
...
+        The mapped dictionary e.g. {"a":"c", "b":"c"}
+
+    """
+    return {
+        list_of_strings[key]: list_of_strings[indices_dict[key]]
+        for key in indices_dict.keys()
+    }
+
+
+def map_installers_same_location_different_id(postcodes, installer_data):
```
athrado: type hint for consistency

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
athrado: Doc string missing?

athrado (commented):
Just a few more comments, mostly regarding consistency in doc strings and type hinting

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: upsi! So sorry about this @sqr00t, I must have somehow (?) changed the location of the function call before adding the last commit. How weird, because all was running fine when I last run it.

I've changed the function call back to where it's supposed to be so that this doesn't happen. 

In short:
- The installers table doesn't originally have the certification body, that's why we do the merge to get the certification body from the installations table;
- However, when we get the missing installers data (from the installations table), we also get the certification body. So the getting the certification body for the original records should be done before we get the missing installers data.

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: see my comment above :)

asf_core_data/pipeline/mcs/process/process_mcs_installations.py
sofiapinto: good shout, completely missed it! :D

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: thanks for the shout :)

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: docstring added now :)

asf_core_data/pipeline/mcs/process/process_mcs_installations.py
sofiapinto: changed :)

sofiapinto:
Hey team @athrado and @sqr00t ,
Thank you so much for your reviews and helpful comments!

Since you've reviewed I have:
1. added Julia's suggestions
3. improved the docstrings in the test file + added additional tests to answer a few of Julia's questions
4. added a new test file for check the outputs of processing installations
5. I've left the code to get lat/long values as is as it might be necessary for a few analysis that only use installers (not for HPMT)
6. sorted the problem with certification body -> in a different way from what you suggested @sqr00t because the function call was basically in the wrong place. Should be sorted now.
7. updated the code to get the latest version of historical installers when processing all MCS data (we currently only have one version, but will be useful when new data comes);
8. added the processing of installers to the `process_and_save_mcs()` function in `generate_mcs_data.py`;
9. while doing 6) I noticed that there were a few problems with using historical data for installers and the quarter data for installations (slight differences between historical installations and quarter installations and the fact that the first few quarters didn't have MCS certification number as a variable, so the values were appearing as missing, when in fact they are not missing). Hence, I have created a new issue #45 that tackles changing the processing of installations to use the historical installations dataset instead of concatenating the quarter installation files.
10. Changed the main README.md to account for the changes.
11. A few small issues came up and I also sorted those: encoding issues, missing postcodes, etc.
12. I've tagged a few functions as legacy and moved them as further down as possible in `process_mcs_utils.py` and `generate_mcs_data.py` as they were previously used but are no longer used, but might be useful not to erase them in case we ever want to run an older analysis and we need these older functions;
13. I've tagged the `process_mcs_installers.py` file as legacy, as we are now working with historical installers data. Didn't delete it because again, we might want to re-run it for an older piece of analysis. I think this file can be excluded from asf-daps though.

@sqr00t I know we've talked about renaming "historical" to "latest" or something similar, but I didn't have time to do it so maybe that can be sorted on asf-daps side.

@sqr00t @athrado, I would ask at least one of you to:
- Run `python3 asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py` and make sure this time it runs without any trouble;
- Run `pytest asf_core_data/pipeline/mcs/test/test_process_historical_mcs_installers.py` - again 17 out 18 tests should pass;
- Run `pytest asf_core_data/pipeline/mcs/test/test_process_mcs_installations.py` - 4/5 tests should pass (1 with warning). one of them doesn't because the full/historical installations file doesn't come with the RHI variables (we can discuss on Monday if we need them and I'll ask Adam to add them back up). It's good to have a test to check for the variables, since now whenever we received new data we can run it to see if the variables come with the same names and if all of them come.
- Run `python3 asf_core_data/pipeline/mcs/generate_mcs_data.py` to make sure it also runs for you. Should take around 30 minutes to run. You need to:
   - have the following folder structure `inputs/EPC/raw_data/2022_Q3_complete` under `asf_core_data`
   - [download this file from S3](https://s3.console.aws.amazon.com/s3/object/asf-core-data?region=eu-west-2&prefix=outputs/EPC/preprocessed_data/2022_Q3_complete/EPC_GB_preprocessed.csv.zip) and place it under `outputs/EPC/preprocessed_data/` and unzipping it. The changes Julia is doing in the other branch will make our life much easier in the future!
- Do let me know if any of the changes look odd, but I don't think you need to look in detail to any changes :) if you do want to do it, you can do so by selecting the changes since you've last reviewed the PR.

Thanks so much for taking the time!

sqr00t:
Thank you for the detailed updates and review instructions :)

1. process_historical_mcs_installers.py -> ran successfully (with the known warning)
2. test_process_historical_mcs_installers.py -> 17/18 passed (test_match_companies_house fails)
3. test_process_mcs_installations -> 4/5 passed (test_raw_columns_exist fails)
4. generate_mcs_data.py -> ran successfully, although your last request in the previous comment has a small typo. The downloaded zip file needs to be under outputs/EPC/preprocessed_data/2022_Q3_complete/

w.r.t. "historical" to "latest" tbh, I was actually thinking the same, happy to sort it out in asf_daps :)

sqr00t (approved):

PR #55, opened 2023-03-21
Removing instances from test account, account for data in multiple sheets, correct EPC_FEAT_SELECTION and a few other fixes

### LINKED ISSUE #52: Remove MCS installations from test account
`"Sharma Solar Thermal Limited 1"` is a test account

### LINKED ISSUE #53: Add NUMBER_HEATED_ROOMS
Add NUMBER_HEATED_ROOMS and remove duplicate of NUMBER_HABITABLE_ROOMS from EPC_FEAT_SELECTION in base_config.py.

Ideally, ceck that both features appear in gold output.

### LINKED ISSUE #54: Account for excel files with multiple sheets
When reading from excel files stored in S3, check if file contains multiple sheets - if so, concatenate data from all sheets.

### LINKED ISSUE #56: Merge corrections
The following merge calls need to be updated.

In asf_core_data/pipeline/preprocessing/feature_engineering.py for get_postcode_coordinates():

`df = pd.merge(df, postcode_coordinates_df, on=postcode_field_name, how="outer")`
--> change _how_ value from 'outer' to 'left'

In asf_core_data/getters/supplementary_data/deprivation/imd_data.py for merge_imd_with_other_set():

 `merged_df = pd.merge(other_df, imd_df, on=["Postcode"])`
--> add kwarg how='left'

And please double-check the merge calls in asf_core_data/pipeline/data_joining/merge_proc_datasets.py
Looks correct to me but just to be safe.

### LINKED ISSUE #57: Getter for processed MCS installations (with or without merged EPC)
Getter for specific versions of processed MCS installations to work with either merged or non-merged to EPC files.

### LINKED ISSUE #58: Creating argument parser that can receive local_data_dir as argument when running scripts

### LINKED ISSUE #59: Change location and name of merged dataset for HPMT consumption
Changes location and name of file that results from running the following script `asf_core_data/pipeline/data_joining/merge_proc_datasets.py`.

The resulting file name is changing from`"/outputs/MCS/merged_epc_mcs_installations_installers.csv"` to `"/outputs/gold/merged_epc_mcs_installations_installers_{date}.csv"` where date is the date when running the file.

### LINKED ISSUE #60: Remove duplicate MCS installations
<1% of historical MCS installations seem to be duplicated.

sofiapinto:
# Description 

This PR fixes a couple of issues:
- #52: Removes instances from accounts used by MCS as test accounts (these instances currently only appear in installations, but we also remove from installers just to be safe);
- #54: Changes the `load_s3_data()` function (in `asf_core_data/getters/data_getters.py`) to account for excel files with multiple sheets. Now, when loading excel files stored in S3, the function check if the file contains multiple sheets. If it contains multiple sheets, it concatenates data from all sheets. This change was needed because the MCS historical installations file is sometimes shared by MCS with data across multiple sheets. Instead of concatenating sheets together locally before uploading raw data to S3 (as we were previously doing), data is now uploaded to S3 exactly as shared by MCS;
- #56: Changes a few merges from outer merge to left merge (a few merges were incorrectly defined as outer merges);
- #57: Creates a getter for any version of processed MCS installations. Previously, we only had a getter for the most recent version of processed MCS installations. This getter was added as Chris needs to use a previous version of MCS installations for one of his projects, because we noticed the latest version of MCS historical installations is missing the `"End User Installation Type"` variable (we have emailed MCS and are waiting to know if this variable can be added back);
- #58: Creates an argument parser that can receive `local_data_dir` as an optional argument when running scripts that load EPC data (as loading EPC from S3 can take very long):
   - Before this update, the script `asf_core_data/pipeline/preprocessing/preprocess_epc_data.py` had to be changed by the user before running (the user needed to set their `LOCAL_DATA_DIR`). Now, the user can set their `LOCAL_DATA_DIR` in the terminal when running the script: e.g. `python3 asf_core_data/pipeline/preprocessing/preprocess_epc_data.py --path_to_data "/Users/anasofiapinto/Documents/ASF data/"`). If `path_to_data` is not provided when running `preprocess_epc_data.py`, the default is to load data from S3.
   - The script to generate the merge dataset for the HPMT can also read EPC data from S3 or locally:
      - from local folder: `python3 asf_core_data/pipeline/data_joining/merge_proc_datasets.py --path_to_data "YOUR/LOCAL/PATH/TO/DATA"`;
      - from S3:   `python3 asf_core_data/pipeline/data_joining/merge_proc_datasets.py`.
- #53: Adds `NUMBER_HEATED_ROOMS` and removes duplicate of `NUMBER_HABITABLE_ROOMS` from `EPC_FEAT_SELECTION` in `asf_core_data/config/base_config.py`. This way, the processed versions of EPC and the merged dataset for HPMT contain the number of heated rooms variable. To account for this change I had to re-run a couple of scripts and upload new versions of processed EPC data to S3 (as described below, hope it makes sense @athrado ):
   - Created a new folder locally to store ASF data ("/Users/anasofiapinto/Documents/ASF data/"). Downloaded the [newest version of raw EPC data (the zipped file)](https://s3.console.aws.amazon.com/s3/object/asf-core-data?region=eu-west-2&prefix=inputs/EPC/raw_data/2022_Q3_complete.zip) and stored it under "/Users/anasofiapinto/Documents/ASF data/inputs/EPC/raw_data/";
   - Ran `python3 asf_core_data/pipeline/preprocessing/preprocess_epc_data.py --path_to_data "/Users/anasofiapinto/Documents/ASF data/"`which generated 3 files under "/Users/anasofiapinto/Documents/ASF data/outputs/EPC/preprocessed/2022_Q3_complete/";
   - Zipped the files individually and uploaded the 6 files [to this folder on S3](https://s3.console.aws.amazon.com/s3/buckets/asf-core-data?region=eu-west-2&prefix=outputs/EPC/preprocessed_data/2022_Q3_complete/&showversions=false);
   - Ran `python3 asf_core_data/pipeline/data_joining/merge_proc_datasets.py --path_to_data "/Users/anasofiapinto/Documents/ASF data/"` which generated the merged dataset for HPMT which was stored [here](https://s3.console.aws.amazon.com/s3/buckets/asf-core-data?region=eu-west-2&prefix=outputs/gold/&showversions=false);
- #59: Changes the name and location of the file containing a merge between installers, installations and EPC data. The file was previous being stored together with the remaining merged versions of EPC and MCS, but given that this is a "special" version used for the HPMT, we're storing it in a different location. We're also adding a date to the file name to allow for versioning;
- #60: <0.1% of lines in MCS historical installations were duplicates - we've removed them.

closes #52 
closes #53 
closes #54
closes #56
closes #57 
closes #58
closes #59
closes #60

## Setup
- checkout to the correct branch: `git checkout 52_remove_mcs_installations_from_test_account`
- Run `make install`;
- Run `direnv allow`;

## Review

Hey @athrado , thank you so much for taking the time to review this PR. I would ask you to please:
- review the changes to the code. Does anything look weird?
- test if running the scripts below works for you
Thanks so much.

Run:
- `python3 asf_core_data/pipeline/data_joining/merge_proc_datasets.py`
If you have any problem due to memory issues, you can do the following instead:
    1. Create a local folder in your computer, [download this file containing EPC processed data from S3](https://s3.console.aws.amazon.com/s3/object/asf-core-data?region=eu-west-2&prefix=outputs/EPC/preprocessed_data/2022_Q3_complete/EPC_GB_preprocessed.csv.zip) and store if under  `YOUR/LOCAL/PATH/TO/DATA/outputs/EPC/preprocessed_data/2022_Q3_complete/`
    2. `python3 asf_core_data/pipeline/data_joining/merge_proc_datasets.py --path_to_data "YOUR/LOCAL/PATH/TO/DATA"`
- Download the [newest version of raw EPC data (the zipped file)](https://s3.console.aws.amazon.com/s3/object/asf-core-data?region=eu-west-2&prefix=inputs/EPC/raw_data/2022_Q3_complete.zip) and stored it under `YOUR/LOCAL/PATH/TO/DATA/inputs/EPC/raw_data/` and run `python3 asf_core_data/pipeline/preprocessing/preprocess_epc_data.py --path_to_data "YOUR/LOCAL/PATH/TO/DATA/"`

Tagging you just so you are aware of the changes: @sqr00t / @Jack-Vines 

---

# Checklist:

- [x] I have refactored my code out from `notebooks/`
- [x] I have tested the code
  - [x] Major functions have docstrings


asf_core_data/getters/data_getters.py
```diff
@@ -229,11 +228,26 @@ def load_s3_data(
...
+                del aux
+        else:  # only one sheet
+            data = pd.read_excel(
+                os.path.join("s3://" + bucket_name, file_name),
+                sheet_name=None,
+                dtype=dtype,
+            )
         if len(data) > 1:
             return data
         else:
```
sqr00t:
```suggestion
        data = pd.read_excel(
            os.path.join("s3://" + bucket_name, file_name),
            sheet_name=None,
            dtype=dtypes,
        )
        if len(data) > 1:
            if isinstance(data, dict) and (len(data.keys()) > 1):
                return pd.concat(
                    data,
                    ignore_index=True,
                )
            else:
                return data
        else:
            return data[list(data.keys())[0]]
```

sqr00t: AFAIK, pd.ExcelFile would eagerly load the excel file anyways (although it's a small file).

I like that the original change is readable, but could it be shortened to something like the above suggestion?

sofiapinto: hey @sqr00t , when you reviewed this file I had not committed all the necessary changes. When you're back from leave, can you please check that the most up to date version of this script works for you? For a bit more context, the changes to this file should fix issue #54

asf_core_data/config/base_config.py
athrado, line 1: Looks good

asf_core_data/getters/data_getters.py
```diff
@@ -229,11 +228,27 @@ def load_s3_data(
...
+
+        # if excel file has data in multiple sheets
+        if len(sheet_names) > 1:
+            data = pd.DataFrame()
+            for sheet in sheet_names:
+                aux = pd.read_excel(
+                    os.path.join("s3://" + bucket_name, file_name),
+                    sheet_name=sheet,
+                    dtype=dtype,
+                )
```
athrado: Looks good, but you can probably get around the aux with something like this (please test/double-check). That should be less computationally expensive than using aux and several concats.

```suggestion
            data = pd.concat([pd.read_excel(
                    os.path.join("s3://" + bucket_name, file_name),
                    sheet_name=sheet,
                    dtype=dtype,
                ) for sheet in sheets])
```

asf_core_data/getters/mcs_getters/get_mcs_installations.py
```diff
@@ -111,3 +111,25 @@ def get_raw_historical_installations_data(
...
         dtype=base_config.raw_historical_installations_dtypes,
     )
+
+
+def get_processed_installations_data(
+    processed_installations_file_name: str,
+) -> pd.DataFrame:
+    """
+    Get a specified version of the processed MCS installation data (both domestic and non-domestic)
+    from S3 (either just MCS or MCS merged with EPC).
```
athrado, line 121: I'm not quite sure what the purpose of this function is, as it just loads the file that is already specified by filename. Personally, I'm not a big believer in having functions with just a one-liner, unless it's a very convoluted one and/or the function is used many many times. Is that the case? Also, you will already have to know which specific file you want to load, and not everyone may know what filenames are available/suitable for their usecase.

I think this function would make more sense if you wouldn't have to give the filename but instead you could specify something like version='most_relevant' and batch='230322' or batch='newest'. Then it would piece together the respective filename (e.g. mcs_installations_epc_most_relevant_230322.csv), which is then loaded. 

Either way, I would specifiy a bit more in the doc string what the different options are for the versions, maybe give some examples. You say 'either merged or not merged with MCS', but aren't there other options, like 'full' vs. 'most relevant'?

asf_core_data/getters/supplementary_data/deprivation/imd_data.py
athrado, line 1: Perfect, thanks a lot for fixing that!

asf_core_data/pipeline/preprocessing/feature_engineering.py
athrado, line 1: Perfect, thanks for fixing this for me!

asf_core_data/pipeline/mcs/process/process_mcs_utils.py
```diff
@@ -223,6 +223,23 @@ def geocode_postcode(data: pd.DataFrame, geodata: pd.DataFrame) -> pd.DataFrame:
...
+    """
+    Drops instances from accounts used for testing by MCS and returns updated data.
+
+    Args:
+        data: installations/installers data
+        company_name_var: name of the variable in data containing the company name
+    Returns:
+        Updated installations/installers data
+    """
+    test_accounts = ["Sharma Solar Thermal Limited 1"]
```
athrado: Very nice!
You use a list instead of a simple string. Do we expect further test accounts? If so, could it make sense to define test_accounts in the config.py as a constant? Just a thought.

asf_core_data/pipeline/data_joining/merge_proc_datasets.py
```diff
@@ -32,7 +33,7 @@ def add_mcs_installations_data(
     epc_df,
     usecols=base_config.MCS_INSTALLATIONS_FEAT_SELECTION,
     bucket_name=base_config.BUCKET_NAME,
-    verbose=False,
+    verbose=True,
```
athrado: Does it make sense to set this o True by default? I think other preprocessing functions have verbose set to False by default, e.g. generate_and_save_mcs(). 
Or maybe we can add a verbose argument to the main pipeling merging_pipeline(verbose=False), which then passes on the verbose argument to the other functions?  Also, define default in doc string.
Just an idea, no need to spend too much time on this.

asf_core_data/pipeline/data_joining/merge_proc_datasets.py
```diff
@@ -222,16 +226,36 @@ def merging_pipeline(
...
-            base_config.EPC_MCS_MERGED_OUT_PATH,
-        )
+    today = date.today().strftime("%y%m%d")
 
-    return epc_mcs_complete
+    # Save final merged dataset
+    data_getters.save_to_s3(
+        base_config.BUCKET_NAME,
+        epc_mcs_complete,
+        base_config.EPC_MCS_MERGED_OUT_PATH.format(today),
```
athrado, line 243: Love that!

asf_core_data/pipeline/data_joining/merge_proc_datasets.py
```diff
@@ -222,16 +226,36 @@ def merging_pipeline(
...
+    )
+
+
+def create_argparser() -> ArgumentParser:
```
athrado, line 247: Always love this stuff! Makes it so much easier to use! Good job!

asf_core_data/pipeline/data_joining/merge_proc_datasets.py
```diff
@@ -222,16 +226,36 @@ def merging_pipeline(
...
+    return parser
 
 
 if __name__ == "__main__":
-    merging_pipeline()
+    parser = create_argparser()
+    args = parser.parse_args()
+
+    path_to_data = args.path_to_data
+    merging_pipeline(path_to_data=path_to_data)
```
athrado: If you want to save a line 

```suggestion
    merging_pipeline(path_to_data=args.path_to_data)
```

asf_core_data/pipeline/mcs/generate_mcs_data.py
```diff
@@ -158,6 +158,48 @@ def generate_and_save_mcs(
     print("Saved in S3: " + most_relevant_epc_path)
 
 
+def get_mcs_installations(epc_version="none", refresh=False):
```
athrado, line 161: Just to double-check: this stayed exactly the same, except that it's no longer labeled LEGACY, right?

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -72,6 +74,8 @@ def basic_preprocessing_of_installations(raw_historical_installations: pd.DataFr
         raw_historical_installations.columns
     )
 
+    raw_historical_installations.drop_duplicates(inplace=True)
```
athrado, line 77: Just checking: So duplicates are dropped if they are exact matches in all fields? Sounds right to me, just wanted to make sure that's intended.

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
athrado, line 1: Looks good!

asf_core_data/pipeline/mcs/process/process_mcs_installations.py
```diff
@@ -151,6 +154,12 @@ def get_processed_installations_data():
         columns=base_config.historical_installations_rename_cols_dict
     )
 
+    installations_data = drop_instances_test_accounts(
+        installations_data, "installer_name"
```
athrado, line 158: Love how this function can be applied to any set (installer or installations) by giving the feature name.

asf_core_data/pipeline/mcs/test/test_process_mcs_installations.py
```diff
@@ -11,6 +11,11 @@
 from asf_core_data.getters.mcs_getters.get_mcs_installations import (
     get_most_recent_raw_historical_installations_data,
 )
+from asf_core_data.pipeline.mcs.process.process_mcs_utils import (
+    drop_instances_test_accounts,
+)
+
```
athrado, line 16: Looks like pytest is no longer needed

asf_core_data/pipeline/mcs/test/test_process_mcs_installations.py
```diff
@@ -21,42 +26,20 @@ def test_raw_columns_exist():
...
+    existing_cols = raw_installations.columns
 
-    assert len([col for col in existing_cols if col not in expected_cols]) == 0
+    for col in expected_cols:
+        if col in ["Installation Type", " Installation Type"]:
+            assert ("Installation Type" in existing_cols) or (
+                " Installation Type" in existing_cols
+            )
+        else:
+            assert col in existing_cols
```
athrado: Just wondering whether this could be done more easily using something like this, but I might be completely wrong:

```suggestion
    expected_cols = [col.strip() for col in expected_cols]
    existing_cols = [col.strip() for col in existing_cols]
    
    assert sorted(set(expected_cols)) == sorted(set(existing_cols))
```

If you only want to strip Installation type, you could do this:

```suggestion
    expected_cols = [col.strip()  if 'Installation Type' in col else col for col in expected_cols]
    existing_cols = [col.strip()  if 'Installation Type' in col else col for col in existing_cols]
    
    assert sorted(set(expected_cols)) == sorted(set(existing_cols))
```

asf_core_data/pipeline/mcs/test/test_process_mcs_installations.py
```diff
@@ -104,16 +87,21 @@ def test_regex_extraction_works():
...
+    ) == len(processed_installations)
 
 
 def test_no_duplicate_lines():
     """
-    Checks that there are no duplicate lines in either raw or processed datasets
+    Checks that there are no duplicate lines in the processed dataset.
     """
-    assert len(raw_installations.drop_duplicates()) == len(raw_installations)
     assert len(processed_installations.drop_duplicates()) == len(
```
athrado, line 98: If processed_installations is a pandas df, it might be better practice to compare the shape instead, either just the rows (.shape[0]) or even the entire shape. Do the number of columns have to match as well?

I've seen you do this on other places as well. Not an expert on tests but I haven't seen people compare Dataframes using length since that's not a very meaningul attribute if you can use shape instead.

```suggestion
    assert processed_installations.drop_duplicates().shape[0] == processed_installations.shape[0])
```

asf_core_data/pipeline/preprocessing/preprocess_epc_data.py
```diff
@@ -293,14 +287,28 @@ def load_and_preprocess_epc_data(
...
-def main():
-    """Main function: Loads and preprocessed EPC data with default settings.
-    Don't forget to set the local data directory."""
+if __name__ == "__main__":
+    parser = create_argparser()
+    args = parser.parse_args()
 
-    # Update local data directory
-    LOCAL_DATA_DIR = "path/to/local_data_dir/"
+    LOCAL_DATA_DIR = args.path_to_data
```
athrado, line 311: So cool!

athrado (commented):
Hey Sofia,

I've just finished reviewing your scripts. It all looks very good, it just left a few comments with a couple of suggestions, but nothing major. 

I was able to run merge_proc_datasets.py without any problems, and it did in fact create a new version in outputs/gold. I'm still running preprocess_epc_data.py with the new arg parser (really loving that!). It's not quite done yet but I don't expect issues beyond this point, and would otherwise let you know.

Thanks a lot for this amazing work! Really enjoyed reviewing the changes and discovering the many different ways you found to make our codebase even better. ;-)

asf_core_data/pipeline/data_joining/merge_proc_datasets.py
sofiapinto: thanks 😊

asf_core_data/pipeline/data_joining/merge_proc_datasets.py
sofiapinto: eheh thanks!

asf_core_data/pipeline/mcs/generate_mcs_data.py
sofiapinto: yep, it was wrongly tagged as legacy!

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: that's exactly what's intended :)

sofiapinto: will just do a few checks to see if these could be representing multiple HPs installed on the same property and being certified on the same they.

asf_core_data/pipeline/preprocessing/feature_engineering.py
```diff
@@ -379,7 +375,7 @@ def get_postcode_coordinates(df, postcode_field_name="POSTCODE"):
     df = data_cleaning.reformat_postcode(df, white_space="remove")
 
     # Merge with location data
-    df = pd.merge(df, postcode_coordinates_df, on=postcode_field_name, how="outer")
+    df = pd.merge(df, postcode_coordinates_df, on=postcode_field_name, how="left")
```
sqr00t, line 378: This function isn't referenced in any pipeline but is similar functionality with asf_core_data.pipeline.mcs.process.process_mcs_utils.
geocode_postcode()

Could you clarify which you'd intend to keep?

sqr00t (approved):
Things look quite solid here, I only wonder where in the pipeline(s) would:

- asf_core_data.pipeline.preprocessing.feature_engineering.get_postcode_coordinates()
- and the functions in imd_data.py

be used.

asf_core_data/pipeline/mcs/test/test_process_mcs_installations.py
sofiapinto: hey @athrado,
- there's another test for the columns, it actually tests more than just the number of columns (tests the names);
- Regarding the use of `len()` vs. `shape[0]`, they do the exact same thing which is returning the number of lines. I don't find anything about using `shape[0]` being best practice instead of using `len()` - could you point me to any resource on this? I actually find `len()` more readable as people might mistake `shape[0]` by `shape[1]`. Here's a [stackoverflow post](https://stackoverflow.com/questions/15943769/how-do-i-get-the-row-count-of-a-pandas-dataframe) with multiple answers comparing `len()` and `shape`.

asf_core_data/pipeline/mcs/process/process_mcs_utils.py
sofiapinto: hey, I didn't know, that's why I created a list but yeah, probably not, so I can use it as a string instead of a list.

asf_core_data/pipeline/mcs/test/test_process_mcs_installations.py
athrado: @sofiapinto Fair enough, it could just be a matter of taste, that's why I said it _might_ be better practice. One of my tutors at uni critised us when using len() with numpy arrays of more than 1-dim, saying it should be avoided when working with tables because it could be unclear whether it's referring to rows or columns or any other dimension. And I always work with .shape myself so I guess I'm just more used to this notation. 

Do what feels right/better readable to you. In fact, len(df.index) might be a marginally faster, just looked it up.

asf_core_data/getters/mcs_getters/get_mcs_installations.py
sofiapinto: hey @athrado ,
I created this function because @ch-williamson needed to access some older MCS processed datasets, as the current version was missing a variable he needed, for a project he is working on at the moment. Hence, instead of downloading the data from S3 and working on it, he can install the asf_core_data package and use this getter function to load a specific version of the data.

I think having "version" and "batch" as arguments is a good idea. What do you think @ch-williamson ? Since you're the one actually using the function atm.

P.S.: In the meantime, MCS have already shared the data with that specific missing variable, so when I re-run the script to re-generate processed MCS data the getter function to load the newest data can be used by Chris instead.

asf_core_data/pipeline/data_joining/merge_proc_datasets.py
sofiapinto: hey, yep I think we can set versbose to False but we need to add the verbose argument to the main pipeline :)

asf_core_data/getters/data_getters.py
sofiapinto: oh I see it now! :) when doing pd.read_excel() for a file with multiple sheets, it will be loaded as a dictionary that we can concatenate. Very smart! Have implemented it - only added a comment to make it obvious for anyone looking at the code. Thanks for the suggestion @sqr00t :)

asf_core_data/getters/data_getters.py
sofiapinto: Have implemented Solomon's suggestion instead, which makes the code a bit shorter and also we only need to load the file once :-)

asf_core_data/pipeline/mcs/test/test_process_mcs_installations.py
sofiapinto: hey @athrado , thanks for the suggestion :) i'll use strip to avoid the if/else statement, but I'll keep the rest as is to compare variable by variable in order to see exactly where it breaks, when it breaks :)

asf_core_data/pipeline/mcs/test/test_process_mcs_installations.py
sofiapinto: that's right :D thanks!

asf_core_data/pipeline/mcs/process/process_mcs_utils.py
sofiapinto: changed to be a string :) we can always change it later

asf_core_data/pipeline/data_joining/merge_proc_datasets.py
sofiapinto: I added verbose as an argument to merging_pipeline() and  as one of the argparser arguments so that it can be defined at runtime.

asf_core_data/pipeline/preprocessing/feature_engineering.py
sofiapinto: good catch @sqr00t ! I am actually not sure why this was created, as I am more familiar with the MCS pipeline than EPC - @athrado where do you use this function?

asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
sofiapinto: update: I did a few checks but will also double check with Adam about this.

asf_core_data/pipeline/preprocessing/feature_engineering.py
athrado: Right, good catch, I meant to use it at the very end after merging all datasets, but the MCS data already has the long/lat added earlier in the pipeline so I ended up forgetting to add it for EPC... I only reformat the postcode.

Maybe you can quickly double-check whether the EPC only records (no MCS match) really don't have the long/lat. If so, we can add a line in merging_pipeline() right after loading epc. 

`feature_engineering.get_postcode_coordinates(prep_epc, postcode_field_name="POSTCODE")`. That should do it!

athrado: We could also add it to the EPC processing pipeline but I found it sometimes takes a bit long and isn't needed for every use case so I removed as a default processing step.

sofiapinto: hey @athrado , thanks for clarifying!

- I have added the missing line of code to `merging_pipeline()` so that we have the coordinates for EPC records;
- @sqr00t I guess to answer your question, for asf_daps you can add only one of the functions to the utils (either the one in `feature_engineering.py` or the one in `process_mcs_utils.py`), since they do roughly the same - the code just needs a bit of adapting;

Issue #73, opened 2023-05-26
Deal with multiple heating systems and fuel types for the same EPC record


sofiapinto:

sofiapinto:
An idea might be to create flag variables, one for each heating system. What is necessary is also project dependent..

Issue #72, opened 2023-05-26
Create unique identifier for properties


sofiapinto:
- [x] For some propeties, we have different EPC entries for the same property sometimes with a UPRN value, other times with missing UPRN. We should fill the missing with the existing UPRN;
- [x] Create a unique identifier that enables uniquely identifying properties even those with missing UPRN.

PR #74, opened 2023-05-26
Enhancements to CONSTRUCTION_AGE_BAND, not allowing for unknowns in UPRN and fix heating features

### LINKED ISSUE #68: Enhance `CONSTRUCTION_AGE_BAND`
Some of `CONSTRUCTION_AGE_BAND` missing values can be inferred base on when the EPC assessment was done and whether it is a new dwelling.

### LINKED ISSUE #70: Issue in `pipeline/preprocessing/feature_engineering.py`
Line 224: `if not (pd.isnull(heating) and isinstance(heating, float)):`

Behaviour if heating is np.nan, this condition would evaluate to False:
`pd.isnull(np.nan)` -> True
`isinstance(np.nan, float)` -> True

Behaviour if heating is None, this condition would evaluate to True,
as "not" (True and False = False) is True

This may be an edge case to address, since the Series "MAINHEAT_DESCRIPTION" of type "object" can contain mixed types of np.nan and None and str.

While integration testing this, I get mismatched lengths at Line 335 as the for loop skips a value when a value in MAINHEAT_DESCRIPTION is np.nan

_Originally posted by @sqr00t in https://github.com/nestauk/asf_core_data/pull/66#discussion_r1196214892_

### LINKED ISSUE #71: `UPRN` with value "unknown"
- Change `standardise_unknowns()` in `asf_core_data/pipeline/preprocessing/data_cleaning.py` to not allow for filling missing with "unknown" for `UPRN`.
- Change code to not identify all properties with missing/unknown UPRN as being the same property

sofiapinto:
# Description 

This PR fixes the following issues in the EPC processing pipeline:
- #71: with the changes we're not allowing for `UPRN` column to take the `“unknown”` value when missing (as it can be misleading since this is an identifier). We also fixed an existing bug when dropping duplicates (when creating `EPC_processed_and_deduplicated.csv`): previously, all EPC entries with `UPRN` missing where considered to be the same property when dropping duplicates, although they in fact represent multiple properties. I tried a temporary fix for this bug, so that when `UPRN` is missing, we now use `ADDRESS1`, `ADDRESS2` and `POSTCODE` to drop duplicates (there might be a better way to this!). It still needs to be fixed further as per issue #72, but that’ll be for a next PR (ideally, we would have a true unique identifier for properties and use it when dropping duplicates). Changes live in `feature_engineering.py` and `epc_data.py`
- #68: currently this variable has approximately 10% missing values. We enhance it by filling the missing values with inspection year if transaction type is “new dwelling”. With this enhancement, missing values decreases to 1%. Changes live in `feature_engineering.py` and `data_cleaning.py`. 
    - Maybe instead of having the `enhance_construction_age_band()` call in `clean_epc_data()` in line 520 maybe it could be called in `clean_CONSTRUCTION_AGE_BAND()` in line 158 – any strong feelings?
    - Can you double check the else if’s in `enhance_construction_age_band()` make sense? I am only wondering about the last ones which depend on `COUNTRY` (I don’t think the year needs to be in the if statement at that point, but I left it there since it helps with readability).
- #70: the `get_heating_features()` function in `feature_engineering.py` had a bug. We fix this bug (by allowing for `None`’s to be treated as `NaN`s) and improve the function by removing the for loop and using `.apply()` and `np.where()` – this speeds up the processing pipeline by a few minutes and improves code readability. I double checked the processed data before and after this change and a few values in the variables created by this function are different. The reason is that sometimes `MAINHEAT_DESCRIPTION` contains multiple applicable values (e.g. “gas” and “electric” for `HEATING_FUEL`) and depending on the order by which we check the presence of these in `MAINHEAT_DESCRIPTION`, we either return one or the other. This is something we should fix in the future (issue #73) by:
    - Considering all the applicable values (e.g. `HEATING_FUEL`=”gas and electric”, `HEATING_SYSTEM`=”boiler and radiators and underfloor heating“);
    - Or by only considering the main value (e.g. main heating fuel for the property is gas, so we consider gas, although the house also has electric heaters);

closes #68
closes #70
closes #71

## Review

Dear @ch-williamson ,

It would be great if you could review the changes to the following scripts:
- `asf_core_data/pipeline/preprocessing/feature_engineering.py`
- `asf_core_data/pipeline/preprocessing/data_cleaning.py`
- `asf_core_data/getters/epc/epc_data.py`

@sqr00t / @Jack-Vines  - tagging you FYI

## Setup

- checkout to the correct branch: `git checkout 70_issue_feature_engineeringpy`
- Run `make install`;
- Run `direnv allow`;

---

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_core_data/getters/epc/epc_data.py
+++ b/asf_core_data/getters/epc/epc_data.py
@@ -39,9 +39,7 @@ def get_cert_rec_files(data_path, dir_name, scotland_data=False):
     """
 
     if str(data_path) == "S3":
-
         if scotland_data:
-
             directories = [
                 Path(f).name
                 for f in data_getters.get_s3_dir_files(path_to_dir=str(dir_name))
@@ -127,9 +125,7 @@ def load_scotland_data(
         data_download.extract_data(data_path / RAW_SCOTLAND_DATA_ZIP)
 
     if scot_usecols is not None:
-
         if v2_batch:
-
             for i, col in enumerate(scot_usecols):
                 if col in base_config.scotland_field_fix_dict.keys():
                     scot_usecols[i] = base_config.scotland_field_fix_dict[col]
@@ -166,7 +162,6 @@ def load_scotland_data(
                 scot_usecols.remove("ENERGY_TARIFF")
 
             if "UPRN" in scot_usecols:
-
                 scot_usecols.remove("UPRN")
                 scot_usecols.append("Property_UPRN")
 
@@ -203,7 +198,6 @@ def load_scotland_data(
         epc_certs["COUNTRY"] = "Scotland"
 
     if v2_batch:
-
         # clean_dict = {"mÂ²": "m2", "Â£": "£", "ï»¿": ""}
         # for col in epc_certs.columns:
         #     for enc_issue in clean_dict.keys():
@@ -270,7 +264,6 @@ def load_england_wales_data(
         add_country_f = True
 
     if subset in [None, "GB", "all"]:
-
         additional_samples = 0
 
         # Splitting samples across nations
@@ -428,7 +421,6 @@ def load_raw_epc_data(
 
     # Get Scotland data
     if subset in ["Scotland", "GB"]:
-
         # data_check = True if subset == "Scotland" else False
 
         epc_scotland_df = load_scotland_data(
@@ -461,9 +453,7 @@ def load_raw_epc_data(
 
     # Merge the two datasets for GB
     elif subset == "GB":
-
         for country in ["Wales", "England"]:
-
             epc_df = load_england_wales_data(
                 data_path=data_path,
                 rel_data_path=wales_england_path,
@@ -675,7 +665,6 @@ def filter_by_year(
 
     # If year is given for filtering
     if year != "all" and year is not None:
-
         if up_to:
             epc_df = epc_df.loc[epc_df["INSPECTION_DATE"].dt.year <= year]
         else:
@@ -685,14 +674,23 @@ def filter_by_year(
     selection_dict = {"first entry": "first", "latest entry": "last"}
 
     if selection in ["first entry", "latest entry"]:
+        epc_df = epc_df.sort_values("INSPECTION_DATE", ascending=True)
 
-        epc_df = (
-            epc_df.sort_values("INSPECTION_DATE", ascending=True)
-            .drop_duplicates(
-                subset=[building_identifier], keep=selection_dict[selection]
-            )
-            .sort_index()
-        )
+        # Dealing with EPC entries with missing UPRN
+        uprn_missing = epc_df[pd.isnull(epc_df["UPRN"])]
+        uprn_missing = uprn_missing.drop_duplicates(
+            subset=["ADDRESS1", "ADDRESS2", "POSTCODE"], keep=selection_dict[selection]
+        ).sort_index()
+
+        # Dealing with EPC entries with known UPRN
+        epc_df = epc_df[~pd.isnull(epc_df["UPRN"])]
+        epc_df = epc_df.drop_duplicates(
+            subset=[building_identifier], keep=selection_dict[selection]
+        ).sort_index()
+
+        # Concatenating datasets together and sorting by inspection_date again
+        epc_df = pd.concat([epc_df, uprn_missing])
+        epc_df = epc_df.sort_values("INSPECTION_DATE", ascending=True)
 
     elif selection is None:
         epc_df = epc_df
--- a/asf_core_data/getters/epc/epc_data.py
+++ b/asf_core_data/getters/epc/epc_data.py
@@ -677,24 +677,26 @@ def filter_by_year(
         epc_df = epc_df.sort_values("INSPECTION_DATE", ascending=True)
 
         # Dealing with EPC entries with missing UPRN
-        uprn_missing = epc_df[pd.isnull(epc_df["UPRN"])]
+        uprn_missing = epc_df[pd.isnull(epc_df[building_identifier])]
         uprn_missing = uprn_missing.drop_duplicates(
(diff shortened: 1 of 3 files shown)


asf_core_data/getters/epc/epc_data.py
```diff
@@ -685,14 +674,23 @@ def filter_by_year(
...
+        epc_df = epc_df.sort_values("INSPECTION_DATE", ascending=True)
+
+        # Dealing with EPC entries with missing UPRN
+        uprn_missing = epc_df[pd.isnull(epc_df["UPRN"])]
+        uprn_missing = uprn_missing.drop_duplicates(
+            subset=["ADDRESS1", "ADDRESS2", "POSTCODE"], keep=selection_dict[selection]
+        ).sort_index()
+
+        # Dealing with EPC entries with known UPRN
+        epc_df = epc_df[~pd.isnull(epc_df["UPRN"])]
```
ch-williamson: Just wondering if this might go wrong when using a different building_identifier, such as BUILDING_REFERENCE_NUMBER? I think this would still work if instead we did
```
uprn_missing = epc_df[pd.isnull(epc_df[building_identifier])]
```
and
```
epc_df = epc_df[~pd.isnull(epc_df[building_identifier])]
```

asf_core_data/getters/epc/epc_data.py
```diff
@@ -685,14 +674,23 @@ def filter_by_year(
...
-            )
-            .sort_index()
-        )
+        epc_df = epc_df.sort_values("INSPECTION_DATE", ascending=True)
+
+        # Dealing with EPC entries with missing UPRN
+        uprn_missing = epc_df[pd.isnull(epc_df["UPRN"])]
+        uprn_missing = uprn_missing.drop_duplicates(
+            subset=["ADDRESS1", "ADDRESS2", "POSTCODE"], keep=selection_dict[selection]
+        ).sort_index()
```
ch-williamson: Do we need to sort here and below if we're sorting by INSPECTION_DATE later?

asf_core_data/getters/epc/epc_data.py
```diff
@@ -685,14 +674,23 @@ def filter_by_year(
...
+        epc_df = epc_df.drop_duplicates(
+            subset=[building_identifier], keep=selection_dict[selection]
+        ).sort_index()
+
+        # Concatenating datasets together and sorting by inspection_date again
+        epc_df = pd.concat([epc_df, uprn_missing])
+        epc_df = epc_df.sort_values("INSPECTION_DATE", ascending=True)
 
     elif selection is None:
         epc_df = epc_df
```
ch-williamson: Just `pass`?

Note that if a selection is passed the dataset gets sorted by INSPECTION_DATE, but if not then it's unsorted - not sure this is a problem necessarily but maybe worth adding to the docstring

asf_core_data/pipeline/preprocessing/feature_engineering.py
```diff
@@ -185,172 +186,197 @@ def map_rating_to_cat(rating):
...
+        ]
+
+        other_hp_expressions = [
+            "heat pump",
+            "pumpa teas",
+            "pwmp gwres",
+            "bwmp gwres",
+            "pympiau gwres",
+            "bympiau gwres",
+        ]
```
ch-williamson: As they're used twice, you could define these up above outside the functions

asf_core_data/pipeline/preprocessing/feature_engineering.py
```diff
@@ -185,172 +186,197 @@ def map_rating_to_cat(rating):
...
+        ):
+            return "boiler, radiator and underfloor"
+
+        elif any(exp in heating for exp in other_heating_system):
+            # If heating system word is found, save respective system type
+            for word in other_heating_system:
+                if word in heating:
+                    return word
+
+        return system_type
```
ch-williamson: `system_type` only ever takes the value `"unknown"`, so could just skip defining it and return `"unknown"` here?

asf_core_data/pipeline/preprocessing/feature_engineering.py
```diff
@@ -433,6 +459,59 @@ def get_building_entry_feature(df, feature):
...
+                elif inspection_year >= 1930:
+                    return "1930-1949"
+                elif inspection_year >= 1919:
+                    return "1900-1929"
+                elif (inspection_year >= 1900) and (country != "Scotland"):
+                    return "1900-1929"
+                elif (inspection_year < 1900) and (country != "Scotland"):
+                    return "England and Wales: before 1900"
+                elif (inspection_year <= 1919) and (country == "Scotland"):
+                    return "Scotland: before 1919"
```
ch-williamson: EPC inspections only began in 2008, so this isn't needed - if it's a new dwelling then the age band will be "2007 onwards"

ch-williamson (commented):
Hi Sofia, looks good - have just suggested some little optimisations. Thanks for fixing this!

On your comment re where cleaning the construction age band should live - it might make more sense for it to be part of `clean_CONSTRUCTION_AGE_BAND`, but I don't feel too strongly about it :)

asf_core_data/getters/epc/epc_data.py
sofiapinto: good point! will fix it!

asf_core_data/pipeline/preprocessing/feature_engineering.py
sofiapinto: good point!

asf_core_data/pipeline/preprocessing/feature_engineering.py
sofiapinto: ahhhh of course, you're absolutely right!

asf_core_data/getters/epc/epc_data.py
sofiapinto: hummm.. I didn't even look at this too much tbh, this is Julia's code. Maybe I can add the sorting to be consistent.

PR #75, opened 2023-05-26
Enhancing installer effective_to

### LINKED ISSUE #69: Improvements to installer `effective_to`
Missing values can be inferred from installation dates together with the date data was received (for now we're just using the latter)

sofiapinto:
# Description 

This PR fixes issue #69 which relates to installer certification dates (`effective_to` is the installer certification end date and `effective_from` is the installer certification start date):
- When `effective_from` is non-missing and `effective_to` is missing, it means that the certification hasn’t ended (that's how the data comes, as raw). Before this PR, we were filling these missing values with the _date_ the data was sent by MCS (which was our best guess). This processing lived in `merge_proc_datasets.py` (the script that creates an outer merge between EPC, MCS installations and MCS installers for the HPMT project), but I now moved it these changes to `process_historical_mcs_installers.py`, where we process installer data, which makes more sense;
- Additionally, I am also enhancing `effective_to` so that we check whether `effective_to` date happens before the last `commissioning_date` for each pair (installer, certification number). If so, then the last commissioning date is actually a better guess for `effective_to` than the date MCS sent us the installer data. This change is being applied to all data, not only when effective_to is initially missing. In a number of cases (small %), we are actually changing the `effective_to` date when it wasn’t initially missing. Is it possible for a heat pump to be commissioned after the certification has ended? Might this be a typo in `effective_to` or shouldn’t we be changing the values of `effective_to` in the first place when the value is not missing? I could have this enhancement only for originally missing values. Let me know what you think :)

closes #69 

## Review

Dear @ch-williamson,

It would be great if you could double check these changes make sense and double check the changes to scripts:
- `asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py`
- `asf_core_data/pipeline/mcs/generate_mcs_data.py`
- `asf_core_data/pipeline/data_joining/merge_proc_datasets.py`

@sqr00t / @Jack-Vines  - tagging you FYI

## Setup

- checkout to the correct branch: `git checkout 69_improvements_installer_effective_to`
- Run `make install`;
- Run `direnv allow`;

---

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_core_data/pipeline/data_joining/merge_proc_datasets.py
+++ b/asf_core_data/pipeline/data_joining/merge_proc_datasets.py
@@ -177,9 +177,6 @@ def add_mcs_installer_data(
         usecols = list(set(usecols + ["company_unique_id", "company_name"]))
 
     newest_hist_inst_batch = data_batches.get_latest_hist_installers()
-    # extracts "20230207" from "mcs_historical_installers_20230207.csv"
-    date = newest_hist_inst_batch.split("_")[-1].split(".csv")[0]
-    date = str(datetime.strptime(date, "%Y%m%d").date())
 
     # Load MCS
     mcs_instllr_data = data_getters.load_s3_data(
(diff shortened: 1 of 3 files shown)


ch-williamson (approved):
Hi Sofia, looks good to me! On the point around replacing `effective_to` dates - I can imagine a situation where an installer is certified until a particular date and the following things happen in sequence:
* MCS pull down their installer data
* The installer renews their accreditation (i.e. the `effective_to` date moves into the future) and then certifies another installation
* MCS pull down their installations data

Maybe we should try and chat to them about it at some point though. If they're not sure we could look into it by e.g. seeing whether the raw `effective_to` dates for these installers change between two batches of data, indicating that they've renewed their accreditation.

sofiapinto:
Thank you @ch-williamson ! I'll double check with MCS :-)

PR #84, opened 2023-06-13
Removing commercial installations from merged dataset and moving function due to circular import

### LINKED ISSUE #78: Remove commercial installations from merged dataset

### LINKED ISSUE #83: Move `enhance_construction_age_band` due to circular import
Move from `feature_engineering.py` to `data_cleaning.py`

sofiapinto:
# Description 

This PR fixes:
- #78: by removing `Commercial` installations from the merged dataset for HPMT, since we're only focusing on `Domestic` installations. Rather than filtering to only keep `Domestic` installations, we filter `Commercial` installations out, so a small percentage of records has missing installation type. I also left the `installation_type` variable in the merged dataset so that the DV/PD team can have the ability to filter records with missing `installation_type`, if they need to.
- #83: by moving the function from `asf_core_data/pipeline/preprocessing/feature_engineering.py` to `asf_core_data/pipeline/preprocessing/data_cleaning.py`. This wasn't giving problems when running some scripts, but it was leading to problems when running `asf_core_data/pipeline/data_joining/merge_proc_datasets.py`.

closes #78 
closes #83 

## Review

Hey @sqr00t , could you please review the scripts with changes? Shouldn't take long. Thanks so much!

## Setup

- checkout to the correct branch: `git checkout 78_remove_commercial_installations_hpmt`
- Run `make install`;
- Run `direnv allow`;

---

### DIFF
--- a/asf_core_data/pipeline/data_joining/merge_proc_datasets.py
+++ b/asf_core_data/pipeline/data_joining/merge_proc_datasets.py
@@ -24,9 +24,9 @@ from argparse import ArgumentParser
 from datetime import date
 import pandas as pd
 import numpy as np
-from datetime import datetime
-from asf_core_data.pipeline.preprocessing import feature_engineering
-from asf_core_data.getters import data_getters
+from asf_core_data.pipeline.preprocessing.feature_engineering import (
+    get_postcode_coordinates,
+)
 
 # ---------------------------------------------------------------------------------
 
@@ -68,7 +68,9 @@ def add_mcs_installations_data(
     )
 
     # Removing commercial installations
-    mcs_df = mcs_df[~mcs_df["installation_type"].str.contains("commercial", case=False)]
+    mcs_df = mcs_df[
+        ~mcs_df["installation_type"].str.contains("commercial", case=False, na=False)
+    ]
 
     mcs_df = install_date_computation.reformat_mcs_date(mcs_df, "commission_date")
     mcs_df.rename(columns={"postcode": "POSTCODE"}, inplace=True)
@@ -251,9 +253,7 @@ def merging_pipeline(
     merged_data = add_mcs_installer_data(merged_data, usecols=mcs_installers_usecols)
 
     # Add coordinates for EPC data
-    merged_data = feature_engineering.get_postcode_coordinates(
-        merged_data, postcode_field_name="POSTCODE"
-    )
+    merged_data = get_postcode_coordinates(merged_data, postcode_field_name="POSTCODE")
 
     today = date.today().strftime("%y%m%d")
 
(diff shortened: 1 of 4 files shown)


asf_core_data/pipeline/data_joining/merge_proc_datasets.py
```diff
@@ -66,6 +67,9 @@ def add_mcs_installations_data(
         dtype={"UPRN": "str", "commission_date": "str"},
     )
 
+    # Removing commercial installations
+    mcs_df = mcs_df[mcs_df["installation_type"] != "Commercial"]
```
sqr00t: This is fine, though I wonder if random spaces or other variations of "Commercial" string would come up.
Could use the following suggestion:

mcs_df = mcs_df[mcs_df["installation_type"].str.contains("Commercial")]

asf_core_data/pipeline/preprocessing/data_cleaning.py
sqr00t, line 1: Shouldn't enhance_construction_age_band() be part of the feature_engineering:get_additional_features()?

Moving this enhancing function this way seems like a bandaid fix for getting merge_proc_datasets to work. Could instead try an explicit import of the only function from feature_engineering (feature_engineering.get_postcode_coordinates()) that is used in merge_proc_datasets?

i.e. from asf_core_data.pipeline.preprocessing.feature_engineering import get_postcode_coordinates

sqr00t (changes requested):
Things to consider:
- Other variations of "Commercial" within installation_type?
- Use an explicit import in merge_proc_datasets.py for get_postcode_coordinates(), instead of importing everything from feature_engineering and moving a feature engineering function into data_cleaning.

asf_core_data/pipeline/data_joining/merge_proc_datasets.py
sofiapinto: It's not a problem so far, but I'll change it to: `mcs_df = mcs_df[~mcs_df["installation_type"].str.contains("commercial", case=False, na=False)]` - this way, it's not case sensitive

asf_core_data/pipeline/preprocessing/data_cleaning.py
sofiapinto: hey @sqr00t - Even when importing the specific functions instead of the whole script, the problem still persists. I've looked further at the functions in `data_cleaning.py` and `feature_engineering.py` and I think the `enhance_construction_age_band()` function might actually might fit `data_cleaning.py` better than `feature_engineering.py`: the work in `enhance_construction_age_band()` is similar to the work in the `standardise_features()` function. Unless you really disagree with this and we find another way for this to work without causing the circular imports problem, I might just leave it at that. We can also always create a new issue later to improve on this, as I'm sure it can be improved.

PR #91, opened 2023-09-01
Enhance UPRN and fix loading for Scotland data

### LINKED ISSUE #72: Create unique identifier for properties

### LINKED ISSUE #86: Changes to Scotland raw field names, encoding and column rows

sofiapinto:
# Description 

This PR fixes:
- #72: prior to this PR, missing `UPRN` was filled with the `BUILDING_REFERENCE_NUMBER` but only for England and Wales, hence `UPRN` had missing values for Scotland. This original enhancement was in script `asf_core_data/getter/epc/epc_data.py`. I removed the enhancement from that script, so that the raw data has missing `UPRN`. We instead now use `BUILDING_REFERENCE_NUMBER` together with address information (`ADDRESS1`, `ADDRESS2` and `POSTCODE`) to enhance `UPRN` during the processing pipeline - changes made to `asf_core_data/pipeline/preprocessing/feature_engineering.py`. Hence, now EPC processed/processed and deduplicated should have no missing `UPRN`.
- #86: I think this has already been fixed by Solomon in asf_daps so no need to change it on your side if loading Scotland data works for you. But in asf_core_data we have problems when loading Scotland data due to column name changes and multiple column headers - so Julia implemented a fix in `load_scotland_data()` in `asf_core_data/getter/epc/epc_data.py` and in `asf_core_data/config/base_config.py`. These changes haven't been committed before, so I'm committing them now.

closes #72
closes #86

## Review

Hey @Jack-Vines and @sqr00t , thanks for reviewing these scripts. Could one of you please double check if the logic looks okay for the changes in `asf_core_data/pipeline/preprocessing/feature_engineering.py`? Thanks a lot!

## Setup
In case you want to run anything (no need to!!!):

- checkout to the correct branch: `git checkout 72_fix_property_id_and_epc_loading`
- Run `make install`;
- Run `direnv allow`;
- Activate the conda environment: `conda activate asf_core_data`;
- Run `python asf_core_data/pipeline/preprocessing/preprocess_epc_data.py` (this will load data from S3, so might take some time). If you have raw data in a local folder you can do `python asf_core_data/pipeline/preprocessing/preprocess_epc_data.py --path_to_data "YOUR/LOCAL/PATH/TO/DATA`. Note that, either way, this will only store the processed data locally.
---

### DIFF
--- a/asf_core_data/getters/epc/epc_data.py
+++ b/asf_core_data/getters/epc/epc_data.py
@@ -357,9 +357,6 @@ def load_england_wales_data(
     if add_country_f:
         epc_certs["COUNTRY"] = subset
 
-    if "UPRN" in epc_certs.columns and "BUILDING_REFERENCE_NUMBER" in epc_certs.columns:
-        epc_certs["UPRN"].fillna(epc_certs.BUILDING_REFERENCE_NUMBER, inplace=True)
-
     if batch in base_config.v0_batches and (
         usecols is None or "BUILDING_REFERENCE_NUMBER" in usecols
     ):
--- a/asf_core_data/getters/epc/epc_data.py
+++ b/asf_core_data/getters/epc/epc_data.py
@@ -185,8 +185,8 @@ def load_scotland_data(
             dtype=dtype,
             low_memory=low_memory,
             usecols=scot_usecols,
-            skiprows=skiprows,  # don't load first row (more ellaborate feature names),
-            encoding="latin-1",
+            skiprows=1,  # don't load first row (more ellaborate feature names),
+            # encoding="latin-1",
         )
         for file in files
     ]
--- a/asf_core_data/pipeline/preprocessing/feature_engineering.py
+++ b/asf_core_data/pipeline/preprocessing/feature_engineering.py
@@ -21,6 +21,7 @@ from hashlib import md5
 
 from asf_core_data.getters.supplementary_data.geospatial import coordinates
 from asf_core_data.pipeline.preprocessing.data_cleaning import reformat_postcode
+from asf_core_data.pipeline.mcs.process.process_mcs_utils import remove_punctuation
 
 # ----------------------------------------------------------------------------------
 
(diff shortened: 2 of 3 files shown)


asf_core_data/getters/epc/epc_data.py
```diff
@@ -185,8 +185,8 @@ def load_scotland_data(
             dtype=dtype,
             low_memory=low_memory,
             usecols=scot_usecols,
-            skiprows=skiprows,  # don't load first row (more ellaborate feature names),
-            encoding="latin-1",
```
sqr00t, line 189: Finally! Ridding of latin-1 encoding was much anticipated!

asf_core_data/pipeline/preprocessing/feature_engineering.py
```diff
@@ -43,6 +44,101 @@
...
+        building_reference: building reference number, e.g. 1234
+        address_info: address information, e.g. "flat 1 abc road_AB2Y8X"
+
+    Returns:
+        A string with concatenated building reference and address info (e.g. "1234_flat 1 abc road_AB2Y8X")
+    """
+    return str(building_reference) + "_" + address_info
+
+
+def enhance_uprn(df: pd.DataFrame) -> pd.DataFrame:
```
sqr00t, line 87: Overrall, I quite like this PR as the changes are specific, and the logic is well commented.

I would extract parts of this out as it's own function (i.e. mapping creation) on the asf_daps side. But I'm happy to merge this PR here. (oops, not the best quoting of the code)

Jack-Vines (approved):
looks good to me too!

sqr00t (approved):

asf_core_data/pipeline/preprocessing/feature_engineering.py
sofiapinto: thanks for review @sqr00t ! Agree with further refactoring the `enhance_uprn()` function, but if you'll do it on asf_daps side, i'll leave it like this there.

PR #106, opened 2023-12-19
105 explore address matching quality

### LINKED ISSUE #105: explore address matching quality
The quality of our current address matching process to join EPC and MCS datasets needs assessment. How effective is it and are there any improvements that we can make?

crispy-wonton:
Addresses #105

Description
Exploratory analysis to assess quality of address matching between MCS and EPC addresses, explore causes of errors, and identify potential optimisation of address score threshold. Files:
1. `get_address_matching_sample.py` 
- takes sample of 1002 non-identical matched addresses for manual review
2. `explore_address_matching_sample.py`
- uses manually reviewed address_matching sample to calculate error rate of address matching and overall error rates in population as well as confidence intervals
- explores patterns in errors and causes of errors
- explores address score distribution across datasets
- explores potential for address score threshold optimisation
3. `requirements.txt` 
- contains additional packages required for running this analysis

Set up:
`explore_address_matching_sample.py`
- activate your core conda environment and then pip install requirements to install additional libraries used
- convert `explore_address_matching_sample.py` to `.ipynb` file and run to see the output tables and visualisations
- run the notebook - it should not require any manual inputs/changes to run

`get_address_matching_sample.py`
- you will need to add your path to local ASF core data to the variable indicated at the beginning of the script, other than that it should work without making any changes.
- please can you run this to check that it works - I think it should work as a .py script but I haven't tested it out
- Note this script takes a while (~10-15 mins) to run because of loading and joining large datasets.
- you do not have to run this before `explore_address_matching_sample.py` as the exploratory notebook pulls all files it uses from S3.

Instructions for reviewer:
- please could you check through the code and check that it all runs ok, and also that the general methodology is appropriate

Please pay special attention to:
- sampling methodology and manual review methodology - do these make sense and are there any changes you would make to make the analysis more robust?
- confidence interval method - I used Wilson based on some reading but definitely not an expert on which method is best so appreciate any inputs on selecting the best one for this purpose. I tested all the methods that statsmodels offers and they all seem to produce roughly the same confidence interval range +/- 1 on either side.
- optimising address score threshold - what's your opinion on the findings?

Checklist:

- [ ] I have tested the code
- [x] I have merged any new changes from `dev`

### DIFF
--- a/asf_core_data/notebooks/explore_address_matching_sample.py
+++ b/asf_core_data/notebooks/explore_address_matching_sample.py
@@ -1,0 +1,485 @@
+# ---
+# jupyter:
+#   jupytext:
+#     cell_metadata_filter: -all
+#     comment_magics: true
+#     text_representation:
+#       extension: .py
+#       format_name: percent
+#       format_version: '1.3'
+#       jupytext_version: 1.15.2
+#   kernelspec:
+#     display_name: asf_core_data_kernel
+#     language: python
+#     name: asf_core_data_kernel
+# ---
+
+# %% [markdown]
+# ### Methodology for manually reviewing address matching sample
+#
+# 1002 matched addresses were manually reviewed to assess whether we can confidently say the match is correct or not. The methodology followed in this process is outlined below:
+# - mcs_address and matched epc_address for each row were manually reviewed.
+# - Addresses that were confidently deemed to be the referring to the same address were assigned a confident_match of 1. Clear mismatches were assigned a confident_match of 0.
+# - Addresses where there was enough difference to create uncertainty were checked in the Royal Mail Address Finder (https://www.royalmail.com/find-a-postcode) by searching the postcode and common words/numeric tokens across both addresses.
+# - If, after checking, there was no ambiguity that mcs_address and epc_address referred to the same address, addresses would be assigned a confident_match of 1. If ambiguity remained, addresses would be assigned a confident_match of 0.
+#
+# #### Illustrative example with fictitious address:
+# mcs_address == 'example farm'\
+# epc_address == 'example farm house'\
+# postcode == 'POST1'
+#
+# We search 'POST1 example farm' in Royal Mail address finder:\
+# Scenario 1: Royal Mail address finder shows the following addresses at 'POST1' postcode:
+# - 'example farm'
+# - 'example farm barn'
+#
+# We cannot be confident which address 'example farm house' refers to so this pairing is given a confident_match of 0. This methodology should be kept in mind when reviewing the results.
+#
+# Scenario 2: Royal Mail address finder shows the following addresses at 'POST1' postcode:
+# - 'example farm'
+#
+# As there is only 1 address at the postcode containing 'example farm', we assume the match is correct and the pairing is given a confident_match of 1.
+#
+# #### Caveats
+# - For ease of reading, we refer in the following analysis to address matches with a confident_match score of 0 as errors. As you can see from the methodology, it is possible that some of these cases are actually true matches but we do not have the data to confidently confirm it. It is therefore possible that this methodology overestimates false positives.
+# - It is also possible that this methodology underestimates false positives, which may occur if addresses have been incompletely or erroneously entered into the original datasets. For example, we may see 'example farm' in both the mcs_address and epc_address fields, but it is possible that one address could really refer to 'example farm' and the other to 'example farm barn'. However, we cannot assess for these possibilities in this analysis. Perhaps it would be possible with geographic data but we do not have that.
+# - This methodology can only account for false positives and not false negatives (i.e. addresses that are true matches but have not been matched by the algorithm).
+
+# %%
+import os
+import pandas as pd
+import numpy as np
+
+from statsmodels.stats.proportion import proportion_confint
+import seaborn as sns
+import matplotlib.pyplot as plt
+
+from sklearn.metrics import roc_curve
+from sklearn.metrics import RocCurveDisplay as rcd
+
+# %% [markdown]
+# ## Set params to retrieve reviewed sample
+
+# %%
+## Params
+
+## Set your sample params
+date = "20231218"
+country_sample_size = 334
+seed = 7
+
+# %% [markdown]
+# ## Read in reviewed address matching sample
+# After manually checking through the sampled rows of matching addresses, we will assess the accuracy of address matching for each country.
+
+# %%
+mcs_epc = pd.read_csv(
+    f"s3://asf-heat-pump-suitability/address-matching/{date}_mcs_epc_address_join_23Q2_prepr_dedupl_231009.csv",
+    keep_default_na=False,
+)
+sample = pd.read_csv(
+    f"s3://asf-heat-pump-suitability/address-matching/{date}_mcs_epc_address_matching_sample_n{country_sample_size*3}_seed{seed}_reviewed.csv",
+    index_col=0,
+)
+
+# %% [markdown]
+# ## Calculate address-matching error rate by country and estimate confidence intervals
+#
+# Here we calculate the error rate (of false positive errors) in address matching for each country in the sample. This represents the error rate (of false positives) of the address matching algorithm because we are only looking at addresses which are non-identical that were matched together. Therefore this tests the effectiveness of the fuzzy matching.
+#
+# We will also calculate the confidence intervals of the error using the Wilson interval. The Wilson interval has been selected as it is recommended for real data applications, https://www.scirp.org/pdf/ojs_2021101415023495.pdf .
+
+# %%
+## Error count in sample
+errors = len(sample[sample["confident_match"] != 1])
+print(f"Total errors in sample: {errors}")
+print(f"Sample size: {len(sample)}")
+print(f"% errors in sample: {round(100 * errors/len(sample), 2)}%")
+
+# %%
+## Calculate error rate by country
+
+sample_country_split = (
+    sample.groupby(["country"])["confident_match"]
+    .value_counts()
+    .reset_index(name="counts")
+)
+sample_country_split.groupby("country")["counts"].sum()
+sample_country_split["error_sample"] = (
+    sample_country_split["counts"] / country_sample_size
+)
+sample_country_split = sample_country_split[
+    sample_country_split["confident_match"] == 0
+]
+
+# %%
+## Calculate confidence intervals for error rates at low and higher bounds in the sample
+## Wilson interval selected as it's recommended for real data applications
+method = "wilson"
+sample_country_split[f"{method}_confidence_interval"] = sample_country_split.apply(
+    lambda row: proportion_confint(
+        count=row["counts"], nobs=country_sample_size, alpha=0.05, method=method
+    ),
+    axis=1,
+)
+
+sample_country_split[f"{method}_lower_error"] = sample_country_split[
+    f"{method}_confidence_interval"
+].str[0]
+sample_country_split[f"{method}_upper_error"] = sample_country_split[
+    f"{method}_confidence_interval"
+].str[1]
+
+# %%
+sample_country_split = sample_country_split[
+    ["country", f"{method}_lower_error", "error_sample", f"{method}_upper_error"]
+]
+sample_country_split
+
+# %% [markdown]
+# The table above shows that the error rate is high. These numbers may also underestimate the true error rate because we cannot account for false negatives in our assessment.
+
+# %% [markdown]
+# ## Estimate overall population error rate
+#
+# Here we will use the error rates we have calculated to estimate the overall number/rate of false positive errors in our population dataset. We will include matches between identical addresses here to estimate overall numbers.
+#
+# We will also calculate confidence intervals for the overall error rate.
+
+# %%
+## View country split in population dataset
+
+country_split = mcs_epc["country"].value_counts(dropna=False).reset_index(name="counts")
+country_split["proportion"] = country_split["counts"] / country_split["counts"].sum()
+country_split
+
+# %%
+## Overall error rate must take into account different country proportions in our population dataset
+## Calculate confidence intervals for population error rate
+
+## Take total count of non-exact matches in population dataset
+country_totals = (
+    mcs_epc.groupby(["country", "exact_match"])
+    .size()
+    .reset_index(name="non_exact_match_counts_popn")
+)
+country_totals = country_totals[country_totals["exact_match"] != 1]
+country_totals = country_totals.drop(columns="exact_match")
+
+## Merge dfs
+popn_error_conf_int = country_totals.merge(
+    sample_country_split, how="left", on="country"
+)
+
+# %%
+## Get number of errors across conf int for each country
+popn_error_conf_int["num_errors_popn"] = (
+    popn_error_conf_int["error_sample"]
+    * popn_error_conf_int["non_exact_match_counts_popn"]
+)
+popn_error_conf_int["num_lower_errors_popn"] = (
+    popn_error_conf_int[f"{method}_lower_error"]
+    * popn_error_conf_int["non_exact_match_counts_popn"]
+)
+popn_error_conf_int["num_upper_errors_popn"] = (
+    popn_error_conf_int[f"{method}_upper_error"]
+    * popn_error_conf_int["non_exact_match_counts_popn"]
+)
+
+## Add in total counts of address matches (including identical) by country across population
+popn_error_conf_int = popn_error_conf_int.merge(
+    mcs_epc.groupby("country").size().reset_index(name="total_counts_popn"),
+    how="left",
+    on="country",
+)
+
+## Get rate of errors across conf int for each country
+popn_error_conf_int["rate_errors_popn"] = (
+    popn_error_conf_int["num_errors_popn"] / popn_error_conf_int["total_counts_popn"]
+)
+popn_error_conf_int["rate_lower_errors_popn"] = (
+    popn_error_conf_int["num_lower_errors_popn"]
+    / popn_error_conf_int["total_counts_popn"]
+)
+popn_error_conf_int["rate_upper_errors_popn"] = (
+    popn_error_conf_int["num_upper_errors_popn"]
+    / popn_error_conf_int["total_counts_popn"]
+)
+
+# %%
+## Split into 2 dfs for display and use
+popn_conf_int_rate = popn_error_conf_int[
+    ["country", "rate_lower_errors_popn", "rate_errors_popn", "rate_upper_errors_popn"]
+]
+popn_conf_int_num = popn_error_conf_int[
+    ["country", "num_lower_errors_popn", "num_errors_popn", "num_upper_errors_popn"]
+]
+
+# %%
+## View rate of predicted errors by country
+popn_conf_int_rate.head()
+
+# %%
+## View number of predicted errors by country
+popn_conf_int_num.head()
+
+# %%
+## Calculate overall confidence interval error rates
+
+popn_error_rates = []
+
+for e in ["num_lower_errors_popn", "num_errors_popn", "num_upper_errors_popn"]:
+    total_est_errors = popn_conf_int_num[e].sum()
+    popn_error_rates.append(round((total_est_errors / len(mcs_epc)) * 100, 2))
+
+# %%
+total_est_errors = popn_conf_int_num["num_errors_popn"].sum()
+print(
+    f"Sample size: {len(sample)}\n\
+% of population sampled: {round(100*(len(sample)/len(mcs_epc)), 2)}%\n\
+Size of population dataset: {len(mcs_epc)}\n\
+Estimated number of false positives in population dataset: {round(total_est_errors)}\n\
+Estimated % of false positives in population dataset: {round((total_est_errors/len(mcs_epc)) *100, 2)}%\n\
+Wilson interval for error rates at 95% confidence: {popn_error_rates[0]}%-{popn_error_rates[2]}%"
+)
+
+# %% [markdown]
+# Here we have calculated the lower and upper bounds of the estimated % of errors in the population level dataset at 95% confidence. However, as many addresses are exact matches, the error rate of the address matching algorithm is actually much higher (as we saw earlier) and requires further investigation.
+
+# %% [markdown]
+# ## Investigate addresses matched to NaN
+# In the sample, there is one erroneous match between an address and a NaN value. Here we will look at why this occurred and how often it occurs across the whole population dataset.
+
+# %%
+mcs_epc_na = mcs_epc.loc[
+    (mcs_epc["mcs_address"] == "") | (mcs_epc["epc_address"] == ""),
+    [
+        "address_1",
+        "address_2",
+        "postcode",
+        "mcs_address",
+        "mcs_postcode",
+        "epc_address",
+        "epc_postcode",
+        "address_score",
+    ],
+]
+mcs_epc_na
+
+# %%
+_epc_values = mcs_epc_na["epc_address"].to_list()
+for val in _epc_values:
+    print(len(val))
+
+# %% [markdown]
+# MCS addresses with NaN or punctuation characters only are converted to a space character (due to this line https://github.com/nestauk/asf_core_data/blob/c7886c041ed08d352a4ec6e088c1e26df90d549c/asf_core_data/pipeline/mcs/process/mcs_epc_joining.py#L51). This space character then matches with short epc_addresses (seemingly 10 or fewer characters) containing a space. This can easily be fixed in the preprocessing pipeline.
+
+# %% [markdown]
+# ## Explore errors in sample for patterns
+#
+# Here we will manually review addresses which have been erroneously matched. We split wrongly matched addresses into alpha-numeric addresses and alpha-only addresses and check for any visible patterns.
+
+# %%
+## Add col to identify addresses with and without numeric components
+sample["numeric_address_component"] = sample.apply(
+    lambda row: 0
+    if (row["mcs_numeric_tokens"] == "set()") & (row["epc_numeric_tokens"] == "set()")
+    else 1,
+    axis=1,
+)
+
+## Get erroneous matches only
+sample_error = sample[sample["confident_match"] != 1]
+
+# %%
+## See which errors have numeric address components
+sample_error_num = sample_error.loc[(sample_error["numeric_address_component"] == 1)]
+sample_error_num
+
+# %% [markdown]
+# From viewing the erroneous addresses with numeric components, it appears there are some patterns to the errors:
+# - Address format: (house_name), number, street, postcode
+# - Not enough information contained in EPC address - e.g. flat/apartment number with no building number/name or street
+
+# %%
+## See which errors don't have numeric components
+sample_error_alpha = sample_error.loc[(sample_error["numeric_address_component"] == 0)]
+sample_error_alpha
+
+# %% [markdown]
+# For addresses without numeric components, it appears there are some patterns to the errors:
+# - Multiple buildings with the same name but different suffix - e.g. ‘example farm’ and ‘example barn’ or ‘example farm’ and ‘example farm cottage’
+# - Similar spelling of house names in the same postcode
+
+# %%
+print(
+    f"Number of errors in addresses in sample with numeric components: {len(sample_error_num)}"
+)
+print(
+    f"% of errors in sample with numeric components: {round(100 * len(sample_error_num)/errors, 2)}%"
+)
+print(
+    f"Number of errors in addresses in sample without numeric components: {len(sample_error_alpha)}"
+)
+print(
+    f"% of errors in sample without numeric components: {round(100 * len(sample_error_alpha)/errors, 2)}%"
+)
+
+# %% [markdown]
+# ## Explore address score distribution
+#
+# We will look at the potential for optimising the address score to see if we can reduce the errors. To do this, we will start by exploring the address score distribution.
+
+# %%
+## See distribution of address matching score in sample
+
+sns.histplot(x="address_score", data=sample, bins=25).set_title(
+    f"Address score distribution in sample data, N={len(sample)}"
+)
+
+# %%
+## Compare to population address matching score distribution
+
+_mcs_epc = mcs_epc[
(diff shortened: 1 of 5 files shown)


asf_core_data/notebooks/explore_address_matching_sample.py
sofiapinto, line 1: To transform into notebook I did:

- ` pip install jupytext`
- `jupytext --to notebook asf_core_data/notebooks/explore_address_matching_sample.py`

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+#
+# We cannot be confident which address 'example farm house' refers to so this pairing is given a confident_match of 0. This methodology should be kept in mind when reviewing the results.
+#
+# Scenario 2: Royal Mail address finder shows the following addresses at 'POST1' postcode:
+# - 'example farm'
+#
+# As there is only 1 address at the postcode containing 'example farm', we assume the match is correct and the pairing is given a confident_match of 1.
+#
+# #### Caveats
+# - For ease of reading, we refer in the following analysis to address matches with a confident_match score of 0 as errors. As you can see from the methodology, it is possible that some of these cases are actually true matches but we do not have the data to confidently confirm it. It is therefore possible that this methodology overestimates false positives.
```
sofiapinto, line 44: When you say the methodology overestimates false positives, you mean your own methodology for reviewing right?

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+# %% [markdown]
+# ## Set params to retrieve reviewed sample
+
+# %%
+## Params
+
+## Set your sample params
+date = "20231218"
+country_sample_size = 334
+seed = 7
```
sofiapinto, line 69: might be worth adding a bit of info on what these mean

asf_core_data/notebooks/explore_address_matching_sample.py
sofiapinto, line 1: it's perhaps worth adding a README explaining that the two notebooks are related and high level information about them - in case we add more notebooks to this folder in the future

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+
+# %%
+mcs_epc = pd.read_csv(
+    f"s3://asf-heat-pump-suitability/address-matching/{date}_mcs_epc_address_join_23Q2_prepr_dedupl_231009.csv",
+    keep_default_na=False,
+)
+sample = pd.read_csv(
+    f"s3://asf-heat-pump-suitability/address-matching/{date}_mcs_epc_address_matching_sample_n{country_sample_size*3}_seed{seed}_reviewed.csv",
+    index_col=0,
+)
```
sofiapinto, line 83: I'm also missing some info on what these two datasets represent.

asf_core_data/notebooks/get_address_matching_sample.py
sofiapinto, line 1: Had to do `export PYTHONPATH=$PWD` before running this as as script, i.e. `python asf_core_data/notebooks/get_address_matching_sample.py`


I also think we need the `os.makedirs()` I added below, as the path does not exist, but otherwise ran fine.

asf_core_data/notebooks/get_address_matching_sample.py
```diff
@@ -0,0 +1,318 @@
...
+
+# %% [markdown]
+# ## Set local_core_data_dir where EPC data is stored
+
+# %%
+## Set your local ASF core data dir
+local_core_data_dir = ""
+
+## Output data will be saved here
+outputs_data_dir = os.path.abspath("../../outputs/data")
```
sofiapinto, line 33: If you do it in the following way, the output will always be saved in the same place, independently of where you're running it from.

And because the folder data doesn't exist, we need to create it.
```suggestion
outputs_data_dir = os.path.join(PROJECT_DIR,"outputs/data")
os.makedirs(outputs_data_dir, exist_ok=True)
```

asf_core_data/notebooks/get_address_matching_sample.py
```diff
@@ -0,0 +1,318 @@
...
+# MCS data: mcs_installations_231009
+
+# %%
+epc_data = pd.read_csv(
+    os.path.join(
+        local_core_data_dir,
+        "outputs/EPC/preprocessed_data/2023_Q2_complete/EPC_GB_preprocessed_and_deduplicated.csv",
+    )
+)
+mcs_data = pd.read_csv("s3://asf-core-data/outputs/MCS/mcs_installations_231009.csv")
```
sofiapinto, line 48: Absolutely no need to change this, but for future reference, you can use the existing getters, e.g. :
```
from asf_core_data.getters import data_getters
from asf_core_data import load_preprocessed_epc_data

data_getters.download_core_data('epc_preprocessed_dedupl', local_core_data_dir, batch="231009")
epc_data = load_preprocessed_epc_data(data_path=local_core_data_dir, version='preprocessed_dedupl')
```

You could even check if the data is stored locally to only download it once. Also to ensure that the version you need is actually stored locally.

asf_core_data/notebooks/get_address_matching_sample.py
```diff
@@ -0,0 +1,318 @@
...
+outputs_data_dir = os.path.abspath("../../outputs/data")
+
+# %% [markdown]
+# ## Read in data
+#
+# EPC data: 2023 Q2 complete, GB dedupl and preprocessed\
+# MCS data: mcs_installations_231009
+
+# %%
+epc_data = pd.read_csv(
```
sofiapinto, line 42: something worth checking:
- I think (?) the matching is done with the non deduplicated version and this might change results slightly in case the same UPRN has different addresses (which I imagine it might happen due to UPRN errors);

asf_core_data/notebooks/get_address_matching_sample.py
```diff
@@ -0,0 +1,318 @@
...
+print(mcs_data.shape)
+
+# %% [markdown]
+# ## Prepare data for address matching
+
+# %%
+epc_df = joining.prepare_epcs(epc_data)
+
+# %%
+mcs_df = joining.prepare_hps(mcs_data)
```
sofiapinto, line 61: I think you told me that `prepare_epcs` and `prepare_hps` are doing slightly different things. Will this have an impact on results?

asf_core_data/notebooks/get_address_matching_sample.py
```diff
@@ -0,0 +1,318 @@
...
+
+# %% [markdown]
+# ## Identify potential address matches
+#
+# Here we calculate an address match score for all candidate matches (i.e. all addresses in the same postcode).
+
+# %%
+matches = joining.form_matching(mcs_df, epc_df)
```
sofiapinto, line 69: I am assuming you're following the exact same methodology as followed in the joining script? - I know you probably couldn't call the exact function used to do the matching because you would miss some variables you need to evaluate the match, but just worth double checking that the methodology followed is exactly the same.

The reason why I'm saying this is that the match is formed in this [script](https://github.com/nestauk/asf_core_data/blob/15ccd754dbc31c2a8ba4d84463cba7ebfe5ebf82/asf_core_data/pipeline/mcs/generate_mcs_data.py) when calling the `join_mcs_epc_data()` function, which uses the `form_matching()` function above.

asf_core_data/notebooks/get_address_matching_sample.py
```diff
@@ -0,0 +1,318 @@
...
+
+# %%
+print(len(matches))
+
+# %% [markdown]
+# ## Join walkthrough
+#
+# Here we will join EPC data to MCS data via address matching using the function used in the asf_core_data pipeline. The function is modified here to retain standardised joining cols and address score. All top matches of EPC to MCS (equivalent to setting all_records=True) will be kept.
```
sofiapinto, line 77: I think this answers my question above, but i'll keep it there just in case :)

sofiapinto: What do you mean by "All top matches of EPC to MCS (equivalent to setting all_records=True) will be kept."?

asf_core_data/notebooks/get_address_matching_sample.py
```diff
@@ -0,0 +1,318 @@
...
+    .droplevel(1)
+    .reset_index()
+    .rename(columns={"index": "level_0"})
+)
+
+# %%
+# Keep standardised address, postcode and numeric tokens columns for reference later
+
+hps = mcs_df.copy()
+epcs = epc_df.copy()
```
sofiapinto, line 105: why do we need to do this?

asf_core_data/notebooks/get_address_matching_sample.py
```diff
@@ -0,0 +1,318 @@
...
+    }
+)
+
+# %%
+# Prepare mcs dataset for left merge
+hps = hps.reset_index()
+
+# %%
+# Join the index-matching df on MCS index to the MCS records
+merged = hps.merge(top_matches, how="left", left_on="index", right_on="level_0")
```
sofiapinto, line 128: I know this probably comes from the original methodology - but are we sure "index" and "level_0" are the same thing? Taking a quick look, it doesn't look like so.

asf_core_data/notebooks/get_address_matching_sample.py
```diff
@@ -0,0 +1,318 @@
...
+    }
+)
+
+# %% [markdown]
+# ## Check output is the same with pipeline function - join datasets
+# Check with the joining function to make sure the join has been correctly replicated.
+
+# %%
+# See matching with existing address-matching function
+mcs_epc_join_function = joining.join_prepared_mcs_epc_data(
```
sofiapinto, line 171: should it be `join_mcs_epc_data()` as this is the one being used [here](https://github.com/nestauk/asf_core_data/blob/15ccd754dbc31c2a8ba4d84463cba7ebfe5ebf82/asf_core_data/pipeline/mcs/generate_mcs_data.py)?

asf_core_data/notebooks/get_address_matching_sample.py
```diff
@@ -0,0 +1,318 @@
...
+    columns=[
+        "mcs_address",
+        "mcs_postcode",
+        "mcs_numeric_tokens",
+        "epc_postcode",
+        "epc_numeric_tokens",
+        "address_score",
+    ]
+)
+assert mcs_epc_join_function.equals(_merged)
```
sofiapinto, line 186: amazing! ⭐️ I see you're checking that you're getting the same thing - and so you answered my question above! I'm really just not sure if using `join_mcs_epc_data()` and `join_prepared_mcs_epc_data()` is the same.

asf_core_data/notebooks/get_address_matching_sample.py
```diff
@@ -0,0 +1,318 @@
...
+        "local_authority": "mcs_local_authority",
+        "COUNTRY": "country",
+    }
+)
+
+# %%
+## Remove rows with no EPC match
+## Any rows with all NaN values across all EPC columns in mcs_epc df will be removed
+epc_cols = epc_df.columns.to_list()
+epc_cols.extend(["epc_address", "epc_postcode", "epc_numeric_tokens"])
```
sofiapinto, line 208: what about removing EPC instances with no match to MCS? Or are those already not in this dataframe?

asf_core_data/notebooks/get_address_matching_sample.py
```diff
@@ -0,0 +1,318 @@
...
+)
+mcs_epc_non_exact = mcs_epc[mcs_epc["exact_match"] != 1]
+
+# %% [markdown]
+# ## Select a sample from the dataset to calculate error rate
+
+# %%
+## We will take 334 rows from each country to create a sample of 1002 address matches
+country_sample_size = 334
+seed = 7
```
sofiapinto, line 254: you selected this one just so you could get the same sample every time you run this script, am I right?

asf_core_data/notebooks/get_address_matching_sample.py
sofiapinto: Understood after going through the whole script :)

asf_core_data/notebooks/get_address_matching_sample.py
sofiapinto, line 1: The methodology looks good! I left a couple of questions, but they're mostly about the matching, not the sampling itself.

In terms of the sampling, there are obviously multiple ways of going about this. Doing it this way, the only thing we're sure is that we have the same number of instances from each country. The sampling does not account for the type of home or other variables, so we might get a highest proportion of a specific type of home and very low of another - but that's the only thing I would point out. I guess you can also check the numbers _a posteriori_ and see if you're happy w/ them.

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+#       format_version: '1.3'
+#       jupytext_version: 1.15.2
+#   kernelspec:
+#     display_name: asf_core_data_kernel
+#     language: python
+#     name: asf_core_data_kernel
+# ---
+
+# %% [markdown]
+# ### Methodology for manually reviewing address matching sample
```
sofiapinto, line 18: Methodology looks good! I guess the only thing I would have done differently would be to tag the "not sure's" as -1 just so you differentiate them from the confident 0s. 

I guess in a way it is important to identify the "not sure's" to answer: is the model doing poorly or is it just doing badly when the data isn't good enough?

asf_core_data/notebooks/get_address_matching_sample.py
crispy-wonton: Yes good point, thank you! I think it should give the same output with either function as `join_mcs_epc_data()` calls `join_prepared_mcs_epc_data()` to do the actual join and then the function also includes the preprocessing steps. But I will update it with `join_mcs_epc_data()` to confirm that the output is the same.

asf_core_data/notebooks/get_address_matching_sample.py
crispy-wonton: It's not necessary in the pipeline but I did it in the notebook to keep those variables so I didnt have to regenerate them every time I made changes as they take a while.

asf_core_data/notebooks/get_address_matching_sample.py
crispy-wonton: Yes I think you are correct because [line 139-144 in `pipeline/mcs/generate_mcs_data.py` ](https://github.com/nestauk/asf_core_data/blob/15ccd754dbc31c2a8ba4d84463cba7ebfe5ebf82/asf_core_data/pipeline/mcs/generate_mcs_data.py#L139) uses default `epcs` parameter, which is `None`, and sets `all_records=True`. Lines [293-294 in `pipleine/mcs/process/mcs_epc_joining.py`](https://github.com/nestauk/asf_core_data/blob/15ccd754dbc31c2a8ba4d84463cba7ebfe5ebf82/asf_core_data/pipeline/mcs/process/mcs_epc_joining.py#L293) show this would set the `epc_version` to `"preprocessed"`.

I will correct this.

asf_core_data/notebooks/get_address_matching_sample.py
crispy-wonton: I think those don't exist in the dataset because everything is left-joined onto the MCS data.

asf_core_data/notebooks/get_address_matching_sample.py
crispy-wonton: Yes correct :)

asf_core_data/notebooks/explore_address_matching_sample.py
crispy-wonton: Yes

asf_core_data/notebooks/get_address_matching_sample.py
crispy-wonton: Thanks, good point!

asf_core_data/notebooks/get_address_matching_sample.py
crispy-wonton: Oh yeah! Thanks for the tip!

asf_core_data/notebooks/get_address_matching_sample.py
crispy-wonton: Yes it should be exactly the same and the output of both joining methodologies is tested below to ensure that they give the same output :)

asf_core_data/notebooks/get_address_matching_sample.py
crispy-wonton: Yes exactly, this is where we test that both methodologies give the same output - as we would expect if the code in the notebook is written correctly.

The 'all top matches' is terminology from [lines 161-163 in the `join_prepared_mcs_epc_data` function](https://github.com/nestauk/asf_core_data/blob/c7886c041ed08d352a4ec6e088c1e26df90d549c/asf_core_data/pipeline/mcs/process/mcs_epc_joining.py#L161). I believe it means that all matched rows for a given address which have the same `address_score` will be kept (i.e. multiple records for the same property). If you set `all_records=False`, you will just get the first occurrence of a given match.

asf_core_data/notebooks/get_address_matching_sample.py
crispy-wonton: Yes I think `level_0` contains numbers representing indices in `hps.index` and `level_1` contains numbers representing indices in `epcs.index`. I've extracted a couple of example indices and checked that they correlate to a sensible row in `epcs` and `hps` and it checks out. I also checked that all values in `level_0` are found in `hps.index` and same with `level_1` and `epcs`.

Is there something else that made you think they were different?

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+
+# %%
+## Calculate error rate by country
+
+sample_country_split = (
+    sample.groupby(["country"])["confident_match"]
+    .value_counts()
+    .reset_index(name="counts")
+)
+sample_country_split.groupby("country")["counts"].sum()
```
sofiapinto, line 107: either print, or remove I guess :)

```suggestion
```

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+sample_country_split["error_sample"] = (
+    sample_country_split["counts"] / country_sample_size
+)
+sample_country_split = sample_country_split[
+    sample_country_split["confident_match"] == 0
+]
+
+# %%
+## Calculate confidence intervals for error rates at low and higher bounds in the sample
+## Wilson interval selected as it's recommended for real data applications
```
sofiapinto, line 117: I don't have a lot of experience with these methods, but after a quick read this one seems to be good for small samples, so sounds good!

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+## Calculate confidence intervals for population error rate
+
+## Take total count of non-exact matches in population dataset
+country_totals = (
+    mcs_epc.groupby(["country", "exact_match"])
+    .size()
+    .reset_index(name="non_exact_match_counts_popn")
+)
+country_totals = country_totals[country_totals["exact_match"] != 1]
+country_totals = country_totals.drop(columns="exact_match")
```
sofiapinto, line 167: so basicaly we're computing the non exact matches for the whole dataset, so that we can then used the estimated proportion wrongly matched to compute how many wrongly matched we will have in the whole dataset, is that it?

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+    popn_error_conf_int["num_errors_popn"] / popn_error_conf_int["total_counts_popn"]
+)
+popn_error_conf_int["rate_lower_errors_popn"] = (
+    popn_error_conf_int["num_lower_errors_popn"]
+    / popn_error_conf_int["total_counts_popn"]
+)
+popn_error_conf_int["rate_upper_errors_popn"] = (
+    popn_error_conf_int["num_upper_errors_popn"]
+    / popn_error_conf_int["total_counts_popn"]
+)
```
sofiapinto, line 207: sorry if i'm missing something here - but why are these proportions different from the above?

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+]
+mcs_epc_na
+
+# %%
+_epc_values = mcs_epc_na["epc_address"].to_list()
+for val in _epc_values:
+    print(len(val))
+
+# %% [markdown]
+# MCS addresses with NaN or punctuation characters only are converted to a space character (due to this line https://github.com/nestauk/asf_core_data/blob/c7886c041ed08d352a4ec6e088c1e26df90d549c/asf_core_data/pipeline/mcs/process/mcs_epc_joining.py#L51). This space character then matches with short epc_addresses (seemingly 10 or fewer characters) containing a space. This can easily be fixed in the preprocessing pipeline.
```
sofiapinto, line 275: this is so weird. how can a space end up matched to a actual existing address?

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+)
+print(
+    f"% of errors in sample with numeric components: {round(100 * len(sample_error_num)/errors, 2)}%"
+)
+print(
+    f"Number of errors in addresses in sample without numeric components: {len(sample_error_alpha)}"
+)
+print(
+    f"% of errors in sample without numeric components: {round(100 * len(sample_error_alpha)/errors, 2)}%"
+)
```
sofiapinto, line 326: this might not mean that there's a bigger tendency for errors to occur in one other than the other, but that your sample has one more than the other

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+
+# %% [markdown]
+# The errors are spread across address scores and not just found in low scores. Let's see a breakdown of the rate of errors across scores.
+
+# %%
+## qcut for equal sized matched address groups
+
+_sample = sample.copy()
+
+_sample["address_score_qbin20"] = pd.qcut(_sample["address_score"], 20)
```
sofiapinto, line 364: cool - have never used the `qcut` function before :)

sofiapinto: perhaps a note on what the 20 means might be good :)

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+_match_rate = _sample.groupby("address_score_bin15").agg({"confident_match": np.mean})
+
+fig, ax = plt.subplots(figsize=(10, 5))
+
+sns.barplot(x=_match_rate.index, y=_match_rate["confident_match"], ax=ax)
+
+ax.set_title(f"Confident Match rate by address score on sample, N={len(_sample)}")
+
+plt.xticks(rotation=90)
+
```
sofiapinto, line 392: little confused. What is the difference between this and the above?

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+sns.histplot(
+    x="address_score",
+    data=sample_error[sample_error["numeric_address_component"] == 0],
+    bins=8,
+).set_title(
+    f"Address score distribution in erroneous address matches for addresses with no numeric components, N={len(sample_error[sample_error['numeric_address_component']==0])}"
+)
+
+# %% [markdown]
+# It appears that erroneous matches of addresses with numeric components tend to have lower address scores and erroneous matches of addresses with no numeric components tend to have higher or very low address scores.
```
sofiapinto, line 419: low count though, hard to take conclusions.

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+    _df["removed_candidates"] = [
+        (sample["address_score"] < threshold).sum() for threshold in _df["threshold"]
+    ]
+    _df["removed_true_matches"] = [
+        (sample[sample["confident_match"] == 1]["address_score"] < threshold).sum()
+        for threshold in _df["threshold"]
+    ]
+    youdens_thresholds_country[country] = _df
+
+youdens_thresholds_country
```
sofiapinto, line 450: perhaps some documentation on what you're doing here would be helpful for future reference. e.g. why are you looking at the Youden's J statistic and what exactly you are doing. At the moment it's a bit hard to follow.

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+
+# %% [markdown]
+# Even when we search for optimum threshold by country or by numeric/alpha address, it looks like there could be a potential optimum address score threshold between 0.84-0.869231.
+#
+# This optimum is based on the following assumptions:
+#
+# - We want to maximise Youden's J-statistic
+# - i.e. sensitivity and specificity are equally important: removing false positives is as important as retaining true positives
+#
+# Based on these results and the confident match rate by address scores assessed above, we should consider increasing the address score threshold at least to 0.74 and potentially to somewhere between 0.77 and 0.87.
```
sofiapinto, line 483: where are these values coming from? you're ouputing the whole df's above, so how are these 0.77 and 0.87 computed?

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+    _df["removed_candidates"] = [
+        (sample["address_score"] < threshold).sum() for threshold in _df["threshold"]
+    ]
+    _df["removed_true_matches"] = [
+        (sample[sample["confident_match"] == 1]["address_score"] < threshold).sum()
+        for threshold in _df["threshold"]
+    ]
+    youdens_thresholds_numeric[key] = _df
+
+youdens_thresholds_numeric
```
sofiapinto, line 473: same as above

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+
+# %% [markdown]
+# Even when we search for optimum threshold by country or by numeric/alpha address, it looks like there could be a potential optimum address score threshold between 0.84-0.869231.
```
sofiapinto, line 476: where are these numbers coming from?

asf_core_data/notebooks/explore_address_matching_sample.py
sofiapinto, line 1: I found this one a bit more difficult to follow than the other one, so maybe adding some more documentation would help. The analysis is truly very thorough - thanks so much for that!

asf_core_data/notebooks/explore_address_matching_sample.py
```diff
@@ -0,0 +1,485 @@
...
+#
+# This optimum is based on the following assumptions:
+#
+# - We want to maximise Youden's J-statistic
+# - i.e. sensitivity and specificity are equally important: removing false positives is as important as retaining true positives
+#
+# Based on these results and the confident match rate by address scores assessed above, we should consider increasing the address score threshold at least to 0.74 and potentially to somewhere between 0.77 and 0.87.
+#
+# However, we lose true positive matches with every increase in the threshold. Also, simply increasing the threshold cannot remove many of the erroneous matches in non-numeric addresses because many of these have very high similarity scores. We will need a different methodology to address these errors. We will explore alternative improvements to address matching in a separate analysis.
```
sofiapinto, line 485: I guess a question I have is: is the best we can get from this model (by doing the above improvements) enough? it's certainly better than we have now, but we potentially need a better address matching model.

sofiapinto (commented):
hey there @crispy-wonton, great work as always!

- very thorough analysis of the errors, although sometimes i found it a bit hard to follow (i think it just needs a bit more documentation)
- I ran the two notebooks/scripts - all good - left 1 or 2 notes about this;
- I mostly left comments/questions for clarification!
- I've also left comments about the sampling methodology, the interval method (although, I didn't have much to add) and about optimising the address score threshold.

asf_core_data/notebooks/explore_address_matching_sample.py
crispy-wonton: Great point for important distinction. And I guess if we have that info we can always change it to a binary score later if necessary but good to differentiate.

asf_core_data/notebooks/explore_address_matching_sample.py
crispy-wonton: Great spot thank you

asf_core_data/notebooks/explore_address_matching_sample.py
crispy-wonton: Great, thanks for looking that up!

asf_core_data/notebooks/explore_address_matching_sample.py
crispy-wonton: yep that's the idea :)

asf_core_data/notebooks/get_address_matching_sample.py
sofiapinto: it's just that I don't see where `level_1`/ `level_0`are coming from - maybe they're computed prior to this?

asf_core_data/notebooks/explore_address_matching_sample.py
crispy-wonton: Yes good point - I'll need to update it so it's not misleading

asf_core_data/notebooks/get_address_matching_sample.py
crispy-wonton: Yes, `prepare_hps` concatenates `address_1` and `address_2` from mcs address fields and then standardises them with by removing punctuation and converting to lower case. `prepare_epcs` only standardises `ADDRESS1` field without concatenating any other address lines. I've now updated the `prepare_epcs` function to use `ADDRESS2` and will review the effect. I think it will have some impact but I think a lot of the initial errors discovered will still be present.

PR #108, opened 2024-07-17
fix datetime issue in installers data

### LINKED ISSUE #107: Issue with datetime variables in installers data

sofiapinto:
When running the MCS-EPC pipeline, I was getting a an error due to variables not being of type datetime. This PR fixed it. 

Fixes #107 
---

Checklist:

- [ ] I have tested the code


asf_core_data/pipeline/mcs/process/process_historical_mcs_installers.py
```diff
@@ -609,6 +608,11 @@ def update_effective_to_date(
         right_on=installations_match_vars,
     )
 
+    installer_data["effective_to"] = pd.to_datetime(installer_data["effective_to"])
+    installer_data["commissioning_date"] = pd.to_datetime(
+        installer_data["commissioning_date"]
+    )
+
```
sqr00t, line 615: a welome addition. I wonder if it would've been better to convert in earlier processing steps of the respective datasets (i.e. on reading in the datasets)?

sqr00t: I did previously encounter issues writing the parquet file as a column of mixed types cannot be written to parquet. So I did add a conversion bit after calling this function in [asf_daps process_mcs_flow.py](https://github.com/nestauk/asf_daps/blob/05dcc4aebb0e1e2ce202605302d9dd65bf330e67/asf_daps/flows/processed/mcs/process_mcs_flow.py#L101-L105)

crispy-wonton: @sqr00t does it make mixed types to convert to datetime if there are NA values?

sqr00t: It should now convert non datetime to pd.NA

crispy-wonton: sorry I meant, does pd.Na and datetime values in the same column mean it would have mixed types and thus would not be able to write to parquet?

sqr00t: There should be no issue. It's fine to have pd.NA and datetime in the same column.

On write, I pass in a PyArrow schema object that checks and can optionally force conversion. pd.NA and datetime in the same column is convertible.

crispy-wonton: awesome thanks for confirming!

crispy-wonton (commented):
Looks fine to me in terms of code :)

The only thing I guess to bear in mind is that there might be some other scripts and functions in other places that expect the date columns to be in string format so they might break with this change and need updating. That being said, unless these columns eventually get renamed to something else, I don't think we use them too much?

